PR #4, opened 2021-05-26
Updates for running on cloud.

jeffmendoza:
Moved operator config to a central package. Keep this in code for now, explore
external config later if needed. Pull app private key from cloud with
go-cloud. Make bot run forever and shut down cleanly with signals. Add files to
build and deploy to AppEngine Flex.


app.yaml
```diff
@@ -0,0 +1,4 @@
+runtime: custom
+env: flex
```
dlorenc, line 2: any reason we need flex vs. cloud run or just a job on k8s?

jeffmendoza: It was quick and simple to get going. Just those two lines, and the single app-deploy command, no other config needed.

Looks like cloud run throttles off the CPU when the app is not serving a request, doesn't work well with background tasks, which we need. They have a feature to send your app a request periodically to solve that, so I could rework the code and setup that request to work on cloud run. Might be an option.

K8s will work, but I just need one VM one pod one process for now. I might go that way because deploying to flex takes forever.

dlorenc (approved):

PR #15, opened 2021-06-25
Log all results, fix override on BP.

jeffmendoza:
Run policies on all repos the app has access to and log results. Actions will
only be taken on enabled repos. Fix override in branch protection polict, and
allow action to be overidden.

### DIFF
--- a/pkg/enforce/enforce.go
+++ b/pkg/enforce/enforce.go
@@ -87,11 +87,10 @@ func EnforceAll(ctx context.Context, ghc *ghclients.GHClients) error {
 			opt.Page = resp.NextPage
 		}
 		for _, r := range repos {
-			if config.IsBotEnabled(ctx, ic, *r.Owner.Login, *r.Name) {
-				err := RunPolicies(ctx, ic, *r.Owner.Login, *r.Name)
-				if err != nil {
-					return err
-				}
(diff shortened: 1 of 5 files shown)


dlorenc (approved):

pkg/enforce/enforce.go
```diff
@@ -87,11 +87,10 @@ func EnforceAll(ctx context.Context, ghc *ghclients.GHClients) error {
 			opt.Page = resp.NextPage
 		}
 		for _, r := range repos {
-			if config.IsBotEnabled(ctx, ic, *r.Owner.Login, *r.Name) {
-				err := RunPolicies(ctx, ic, *r.Owner.Login, *r.Name)
-				if err != nil {
-					return err
-				}
+			enabled := config.IsBotEnabled(ctx, ic, *r.Owner.Login, *r.Name)
```
inferno-chromium, line 90: Why do you need to check enabled bit seperately, why not do this inside runpolicies that gets all these args.

inferno-chromium (approved):

inferno-chromium (commented):
good to clean those github action lint failures on return vals not checked.

pkg/enforce/enforce.go
jeffmendoza: It was just for ease of unit testing.

PR #19, opened 2021-07-22
New policies

jeffmendoza:
Add 3 new policies to allstar.

- Security.md to verify that exists.
- Outside collaborator, to prevent admins outside an org. 
- Binary Artifacts, the first policy adapted from a scorecard check.

Note: moving to an older version of go-github to match scorecard until https://github.com/google/go-github/pull/2016 is released.


go.mod
```diff
@@ -3,13 +3,17 @@ module github.com/ossf/allstar
...
-	github.com/google/go-querystring v1.1.0 // indirect
+	github.com/google/go-github/v29 v29.0.3 // indirect
+	github.com/google/go-github/v32 v32.1.0
 	github.com/gregjones/httpcache v0.0.0-20190611155906-901d90724c79
-	github.com/niemeyer/pretty v0.0.0-20200227124842-a10e7caefd8e // indirect
-	github.com/rs/zerolog v1.22.0 // indirect
-	gocloud.dev v0.23.0 // indirect
-	gopkg.in/check.v1 v1.0.0-20200227125254-8fa46927fb4f // indirect
-	gopkg.in/yaml.v2 v2.2.8
+	github.com/ossf/scorecard v1.2.1-0.20210722153731-89c8e2af3131
```
azeemshaikh38, line 12: FYI after https://github.com/ossf/scorecard/pull/716 needs to be updated to `scorecard/v2`

pkg/policies/binary/binary.go
```diff
@@ -0,0 +1,212 @@
...
+	oc, rc := getConfig(ctx, c, owner, repo)
+	enabled := config.IsEnabled(oc.OptConfig, rc.OptConfig, repo)
+	log.Info().
+		Str("org", owner).
+		Str("repo", repo).
+		Str("area", polName).
+		Bool("enabled", enabled).
+		Msg("Check repo enabled")
+
+	repoClient := githubrepo.CreateGithubRepoClient(ctx, c)
```
azeemshaikh38, line 117: Use `RunScorecards` fn directly?

https://github.com/ossf/scorecard/blob/main/pkg/scorecard.go#L72

pkg/policies/binary/binary.go
```diff
@@ -0,0 +1,212 @@
...
+	}
+	// TODO, likely this should be a "scorecard" policy that runs multiple checks
+	// here, and uses config to enable/disable checks.
+	res := checks.BinaryArtifacts(cr)
+	if res.Error2 != nil {
+		return nil, res.Error2
+	}
+
+	var notify string
+	if res.Score < checker.MaxResultScore {
```
azeemshaikh38, line 140: Should the value we compare to (`MaxResultScore`) be part of the policy?

jeffmendoza: Yes! I see that binary artifacts is just using max/min right now, so went with that for now:
https://github.com/ossf/scorecard/blob/main/checks/binary_artifact.go#L41-L45

That said, I think this policy should be expanded so that the operator can list in config which checks to run and which score is considered a "pass" for each check

pkg/policies/security/security_test.go
```diff
@@ -0,0 +1,167 @@
...
+	"github.com/google/go-github/v32/github"
+	"github.com/ossf/allstar/pkg/config"
+	"github.com/ossf/allstar/pkg/policydef"
+)
+
+var getContents func(context.Context, string, string, string,
+	*github.RepositoryContentGetOptions) (*github.RepositoryContent,
+	[]*github.RepositoryContent, *github.Response, error)
+
+type mockRepos struct{}
```
azeemshaikh38, line 34: Update to:

```
type mockRepos struct {
  ctx context.Context
  owner, repo, path string
  opts *github.RepositoryContentGetOptions
}

func (m mockRepos) GetContents(...) (...) {
  // Logic operating on m.ctx, m.owner, m.repo etc.
}
```

And then, you just initialize a new instance of mockRepos in the test: 

```
m := mockRepos {
 ctx: ...,
 repo: ...,
}
```

pkg/policies/security/security_test.go
```diff
@@ -0,0 +1,167 @@
...
+					}, nil, nil, nil
+				} else {
+					return nil, nil, &github.Response{Response: &http.Response{StatusCode: http.StatusNotFound}}, errors.New("Not Found")
+				}
+			}
+			res, err := check(context.Background(), mockRepos{}, nil, "", "thisrepo")
+			if err != nil {
+				t.Fatalf("Unexpected error: %v", err)
+			}
+			if !reflect.DeepEqual(res, &test.Exp) {
```
azeemshaikh38: Consider using `github.com/google/go-cmp/cmp` here.

pkg/policies/security/security.go
```diff
@@ -0,0 +1,206 @@
...
+	"github.com/ossf/allstar/pkg/config"
+	"github.com/ossf/allstar/pkg/config/operator"
+	"github.com/ossf/allstar/pkg/policydef"
+
+	"github.com/google/go-github/v32/github"
+	"github.com/rs/zerolog/log"
+)
+
+const configFile = "security.yaml"
+const polName = "SECURITY.md"
```
azeemshaikh38, line 33: The policy name being `SECURITY.md` is intentional or typo? Maybe something like `Security`?

pkg/policies/security/security.go
```diff
@@ -0,0 +1,206 @@
...
+	log.Info().
+		Str("org", owner).
+		Str("repo", repo).
+		Str("area", polName).
+		Bool("enabled", enabled).
+		Msg("Check repo enabled")
+
+	fc, _, rsp, err := rep.GetContents(ctx, owner, repo, filePath, nil)
+	if err != nil {
+		if rsp != nil && rsp.StatusCode == http.StatusNotFound {
```
azeemshaikh38, line 113: Should this be outside `if err != nil` block? If `err != nil` the response returned should be invalid/irrelevant and to be ignored?

pkg/policies/policies.go
```diff
@@ -17,13 +17,19 @@
...
 import (
+	"github.com/ossf/allstar/pkg/policies/binary"
 	"github.com/ossf/allstar/pkg/policies/branch"
+	"github.com/ossf/allstar/pkg/policies/outside"
+	"github.com/ossf/allstar/pkg/policies/security"
 	"github.com/ossf/allstar/pkg/policydef"
 )
 
 // GetPolicies returns a slice of all policies in Allstar.
 func GetPolicies() []policydef.Policy {
```
azeemshaikh38, line 28: Not needed in this PR, but might be useful to consider registering policies like this: https://github.com/ossf/scorecard/blob/35267c251496135a0a5e9097470583f6c447b68c/checks/all_checks.go#L22

In case users want to run only a smaller list of policies it'll be useful.

pkg/policies/outside/outside_test.go
```diff
@@ -0,0 +1,175 @@
...
+
+	"github.com/google/go-github/v32/github"
+	"github.com/ossf/allstar/pkg/config"
+	"github.com/ossf/allstar/pkg/policydef"
+)
+
+var listCollaborators func(context.Context, string, string,
+	*github.ListCollaboratorsOptions) ([]*github.User, *github.Response, error)
+
+type mockRepos struct{}
```
azeemshaikh38, line 30: Similar comment here as the other `_test.go` file for `mockRepos`.

azeemshaikh38 (dismissed):

pkg/policies/security/security.go
jeffmendoza: I debated this. GitHub calls having a SECURITY.md file a "Security Policy", so that would make sense to call this policy "Security". However, in Allstar we are calling it a "Security Policy App" are categorizing all the policies as "Security Policies", so I need to differentiate this one.

pkg/policies/policies.go
jeffmendoza: Yes, I'm familiar with this pattern. While useful, I think it reduces clarity, so I'd only want to use if with good reason. At the user level: which policies are run is set by configuration. If we do get more app operators that wish to exclude policies at compile time it would make sense.

pkg/policies/binary/binary.go
jeffmendoza: Looking at `RunScorecards` I think it includes some code that I don't necessarily want to run in Allstar. Lets continue the discussion of what would be a nice clean way to interface and develop a plan.

pkg/policies/security/security_test.go
jeffmendoza: Updated ptal.

jeffmendoza:
@azeemshaikh38 @laurentsimon Can I get this re-approved ( I have dismiss stale turned on )

azeemshaikh38 (approved):

azeemshaikh38:
> @azeemshaikh38 @laurentsimon Can I get this re-approved ( I have dismiss stale turned on )

done.

PR #25, opened 2021-08-01
🌱  Implemented golangci checks similar to scorecard

naveensrinivasan:
* Included golangci setting similar to [scorecard](https://github.com/ossf/scorecard/blob/main/.golangci.yml)
  * Included a makefile to run the linter
  * Included tools.go for external dependencies.

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -1,0 +1,154 @@
+---
+run:
+  concurrency: 6
+  deadline: 5m
+issues:
+  new-from-rev: ""
+  include:
+    # revive `package-comments` and `exported` rules.
+    - EXC0012
+    - EXC0013
+    - EXC0014
+    - EXC0015
+linters:
+  disable-all: true
+  enable:
+    - asciicheck
+    - bodyclose
+    - deadcode
+    - depguard
+    - dogsled
+    - dupl
+    - errcheck
+    - errorlint
+    - exhaustive
+    - exportloopref
+    - gci
+    - gochecknoinits
+    - gocognit
+    - goconst
+    - gocritic
+    - gocyclo
+    - godot
+    - godox
+    - goerr113
+    - gofmt
+    - gofumpt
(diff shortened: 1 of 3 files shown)


jeffmendoza:
Thanks for this, though I'm not excited about it. I like to keep things simple where we can. For example, the current job config is just a few lines:
```
    - uses: golangci/golangci-lint-action@v2
      with:
        version: v1.41.1
```
The lint config is way too verbose
```
linters:
  disable-all: true
  enable:
...
```
I'd rather only list the additional non-default linters that we want to enable. The file should only contain the necessary lines.

As for the Makefile, I prefer just using `go` commands, and devs being able to `go install` the lint tools on their own. A developers.md file with the go commands to install / build / test is preferred to Makefiles and the tools.go

naveensrinivasan:
> Thanks for this, though I'm not excited about it. I like to keep things simple where we can. For example, the current job config is just a few lines:
> 
> ```
>     - uses: golangci/golangci-lint-action@v2
>       with:
>         version: v1.41.1
> ```

With the present setup, anyone would have to push to get the lint issues. This avoids that.

> 
> The lint config is way too verbose
> 
> ```
> linters:
>   disable-all: true
>   enable:
> ...
> ```
> 
> I'd rather only list the additional non-default linters that we want to enable. The file should only contain the necessary lines.
> 
> As for the Makefile, I prefer just using `go` commands, and devs being able to `go install` the lint tools on their own. A developers.md file with the go commands to install / build / test is preferred to Makefiles and the tools.go

The issue with the approach of installing tools based on developers.md is that it is not pinning the version that was required to test/lint the code. The `tools.go` provides a way to pin dependencies on these external tools.  

I understand if you still have strong opinions.

jeffmendoza:
Thanks for the reasoning and understanding, I've opened #39 and #40 to cover this.

PR #34, opened 2021-08-18
Bump resources

jeffmendoza:


inferno-chromium (approved):
Can go even higher in memory and disk for future use as well.

PR #50, opened 2021-09-16
Handle branch protection on private repositories.

### LINKED ISSUE #37: Handle BP on free-tier private repos
Branch Protection is not enabled on free-tier private repos: https://docs.github.com/en/github/administering-a-repository/defining-the-mergeability-of-pull-requests/about-protected-branches
GitHub responds with:
`GET https://api.github.com/repos/org/repo/branches/master/protection: 403 Upgrade to GitHub Pro or make this repository public to enable this feature. []`
Currently we error out of the whole installation/org.

Change to fail BP policy, notify text should include the message from GitHub "Upgrade to GitHub Pro or make this repository public to enable this feature."

jeffmendoza:
Fixes #37

### DIFF
--- a/pkg/policies/branch/branch.go
+++ b/pkg/policies/branch/branch.go
@@ -217,7 +217,7 @@ func check(ctx context.Context, rep repositories, c *github.Client, owner,
 				pass = false
 				text = text + "Branch Protection enforcement is configured in Allstar, however Branch Protection is not available on this repository. Upgrade to GitHub Pro or make this repository public to enable this feature.\n" +
 					"See: https://docs.github.com/en/repositories/configuring-branches-and-merges-in-your-repository/defining-the-mergeability-of-pull-requests/about-protected-branches for more information.\n" +
-					"Alternatively, disable Branch Protection policy enforcement for this repository in Allstar configuration."
+					"If this is not feasible, then disable Branch Protection policy enforcement for this repository in Allstar configuration."
 				ds[b] = details{}
 				break
 			}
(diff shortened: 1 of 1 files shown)


pkg/policies/branch/branch.go
```diff
@@ -212,6 +212,15 @@ func check(ctx context.Context, rep repositories, c *github.Client, owner,
 				ds[b] = details{}
 				continue
 			}
+			if rsp != nil && rsp.StatusCode == http.StatusForbidden {
+				// Protection not available
+				pass = false
+				text = text + "Branch Protection enforcement is configured in Allstar, however Branch Protection is not available on this repository. Upgrade to GitHub Pro or make this repository public to enable this feature.\n" +
+					"See: https://docs.github.com/en/repositories/configuring-branches-and-merges-in-your-repository/defining-the-mergeability-of-pull-requests/about-protected-branches for more information.\n" +
+					"Alternatively, disable Branch Protection policy enforcement for this repository in Allstar configuration."
```
inferno-chromium: Can remove this alternatively line, it is bad recommendation?

inferno-chromium (dismissed):

pkg/policies/branch/branch.go
jeffmendoza: If folks don't want to pay for GitHub pro, they can't setup branch protection on private repos. So they probably don't want Allstar configured to bug them to turn it on.

inferno-chromium: ok, then can frame like
"If this is not feasible, then disable ..."

Alternatively makes it think it is ok to do that. so slight language change, then can keep line. thanks for explaining

jeffmendoza: Yes, sounds great. Applied. Please re-approve.

inferno-chromium (approved):

PR #51, opened 2021-09-16
Update GitHub client version.

jeffmendoza:
Before we were in lock-step with Scorecards in order to share clients. Now,
go-github has added a method to get the internal http client, therefore we can
create new GitHub clients on the fly to convert major versions. See binary.go
for conversion.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -8,11 +8,11 @@ require (
 	github.com/google/go-cmp v0.5.6 // indirect
 	github.com/google/go-github/v29 v29.0.3 // indirect
 	github.com/google/go-github/v32 v32.1.0
+	github.com/google/go-github/v39 v39.0.0 // indirect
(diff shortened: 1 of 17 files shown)


go.mod
```diff
@@ -8,11 +8,11 @@ require (
 	github.com/google/go-cmp v0.5.6 // indirect
 	github.com/google/go-github/v29 v29.0.3 // indirect
 	github.com/google/go-github/v32 v32.1.0
```
inferno-chromium, line 10: i think we need to clean these older version sometime.

inferno-chromium (approved):
Thanks!

go.mod
jeffmendoza: 29 is used indirectly somewhere as a sub-dep. 32 is used by scorecards. I'll make an issue there.

Issue #43, opened 2021-08-31
Organization wide SECURITY.md file does not fulfill SECURITY.md policy


josephperrott:
If an organization has an organization level `SECURITY.md` file defined as outline in [About default community health files](https://docs.github.com/en/communities/setting-up-your-project-for-healthy-contributions/creating-a-default-community-health-file), I would expect that this would satisfy the SECURITY.md policy check.

Instead, an [issue](https://github.com/angular/dev-infra/issues/192) was created with a link to create this policy, but this link is actually just to our [policy](https://github.com/angular/dev-infra/security/policy) itself, which is defined [here](https://github.com/angular/.github/blob/master/SECURITY.md).

jeffmendoza:
Thanks for pointing this out. It looks like setting the org-wide security policy does not make a phantom SECURITY.md "appear" in the repo, which is what this policy looks for. Looks like just the security tab is updated in the UI. We will have to find a way to detect this in the API, or if that does not exist, just re-implement GitHub's logic and check the org-level `.github` repo.

josephperrott:
From a quick look it looks like using Github's v4 GraphQL API you can request this information directly.

The [Repository](https://docs.github.com/en/graphql/reference/objects#repository) object has an attribute `isSecurityPolicyEnabled`.  Personally, I don't have enough experience with go to just jump into this repo and try to make it happen.  But it seems to be pretty straight forward if you would like to leverage the Github GraphQL API.

azeemshaikh38:
@jeffmendoza FWIW Scorecard already handles the case of scanning the `.gtihub` org repo for security policy file - https://github.com/ossf/scorecard/blob/main/checks/security_policy.go. Maybe consider calling the Scorecard SecurityPolicy check directly?

jeffmendoza:
Some more details from #46
> github docs:
https://docs.github.com/en/code-security/getting-started/adding-a-security-policy-to-your-repository

> Show SECURITY.md can be under root, docs, or .github

jeffmendoza:
Fixed

PR #52, opened 2021-09-16
Update security policy check.

jeffmendoza:
Now checks GitHub's api for policy which looks in `.github` and `docs`
sub-directories, as well as the org-level `.github` repository. This requires
using the v4 graphql api.

### DIFF
--- a/pkg/policies/security/security.go
+++ b/pkg/policies/security/security.go
@@ -75,6 +75,10 @@ func init() {
 	configFetchConfig = config.FetchConfig
 }
 
+type v4client interface {
+	Query(context.Context, interface{}, map[string]interface{}) error
+}
+
 // Security is the SECURITY.md policy object, implements policydef.Policy.
 type Security bool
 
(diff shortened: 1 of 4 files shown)


jeffmendoza:
Fixes #43

pkg/policies/security/security.go
```diff
@@ -114,29 +112,27 @@ func check(ctx context.Context, rep repositories, c *github.Client, owner,
...
-				NotifyText: "SECURITY.md not found.\n" + fmt.Sprintf(notifyText, owner, repo),
-				Details: details{
-					Exists: false,
-					Empty:  true,
-				},
-			}, nil
-		}
+	var q struct {
+		Repository struct {
+			SecurityPolicyUrl       string
```
azeemshaikh38, line 116: Should these be `githubv4.String` and `githubv4.Boolean`?

pkg/policies/security/security.go
```diff
@@ -27,6 +26,7 @@ import (
 
 	"github.com/google/go-github/v39/github"
 	"github.com/rs/zerolog/log"
+	"github.com/shurcooL/githubv4"
```
azeemshaikh38, line 29: Remove `filePath`?

azeemshaikh38 (dismissed):
LGTM. FYI, Scorecard decided against this approach since it limits checking for security policy in non-GitHub VCS.

laurentsimon (dismissed):

pkg/policies/security/security.go
jeffmendoza: https://github.com/shurcooL/githubv4#scalar-types
> However, depending on how you're planning to use the results of your query, it's often more convenient to use other Go types.
> The encoding/json rules are used for converting individual JSON-encoded fields from a GraphQL response into Go values. See https://godoc.org/encoding/json#Unmarshal for details. The json.Unmarshaler interface is respected.
> That means you can simplify the earlier query by using predeclared Go types:

Using `string` is more convenient, here, don't need to check the pointer for nil, etc.

pkg/policies/security/security.go
jeffmendoza: Fixed.

jeffmendoza:
> LGTM. FYI, Scorecard decided against this approach since it limits checking for security policy in non-GitHub VCS.

Yeah, makes sense, different use cases.

Updated, PTAL.

pkg/policies/security/security.go
azeemshaikh38: SGTM.

azeemshaikh38 (approved):

PR #54, opened 2021-09-19
Opt out private/public repos by optOutPrivateRepos/optOutPublicRepos

five510:
## Changes
- Opt out private/public repos by optOutPrivateRepos/optOutPublicRepos
- Add how to opt out by using optOutPrivateRepos/optOutPublicRepos in opt-out.md

### DIFF
--- a/opt-out.md
+++ b/opt-out.md
@@ -7,19 +7,12 @@ To determine if Allstar is configured in the opt-out strategy, there will be a r
     optConfig:
       optOutStrategy: true
 
-To opt-out specific repository, submit a PR to that `.allstar` repo, and add the name of your repository to the opt-out list. ex:
+To opt-out, submit a PR to that `.allstar` repo, and add the name of your repository to the opt-out list. ex:
 
     optConfig:
       optOutStrategy: true
       optOutRepos:
       - my-repo-name-here
-
-To opt-out all private/public repositories, submit a PR to that `.allstar` repo, and add `optOutPrivateRepos` or `optOutPublicRepos`. ex:
-
-    optConfig:
-      optOutStrategy: true
-      optOutPrivateRepos: true
-      optOutPublicRepos: false
 
 ### With repo-override
 
--- a/opt-out.md
+++ b/opt-out.md
@@ -7,12 +7,19 @@ To determine if Allstar is configured in the opt-out strategy, there will be a r
     optConfig:
       optOutStrategy: true
 
-To opt-out, submit a PR to that `.allstar` repo, and add the name of your repository to the opt-out list. ex:
+To opt-out specific repository, submit a PR to that `.allstar` repo, and add the name of your repository to the opt-out list. ex:
 
     optConfig:
       optOutStrategy: true
       optOutRepos:
       - my-repo-name-here
+
+To opt-out all private/public repositories, submit a PR to that `.allstar` repo, and add `optOutPrivateRepos` or `optOutPublicRepos`. ex:
+
+    optConfig:
+      optOutStrategy: true
+      optOutPrivateRepos: true
+      optOutPublicRepos: false
 
 ### With repo-override
 
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -112,7 +112,11 @@ type repositories interface {
 
 // IsEnabled determines if a repo is enabled by interpreting the provided
 // org-level and repo-level OptConfigs.
-func IsEnabled(ctx context.Context, o OrgOptConfig, r RepoOptConfig, rep repositories, owner, repo string) (bool, error) {
+func IsEnabled(ctx context.Context, o OrgOptConfig, r RepoOptConfig, c *github.Client, owner, repo string) (bool, error) {
+	return isEnabled(ctx, o, r, c.Repositories, owner, repo)
+}
+
+func isEnabled(ctx context.Context, o OrgOptConfig, r RepoOptConfig, rep repositories, owner, repo string) (bool, error) {
 	var enabled bool
 
 	gr, _, err := rep.Get(ctx, owner, repo)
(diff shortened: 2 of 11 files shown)


jeffmendoza:
Thank you for looking into this! I think it would be good for the config to be a part of `OrgOptConfig`. We want each individual policy to be easily turned on/off for all public/private repos. For example Allstar could be used on all repos, but the SECURITY.md policy only enabled on public repos.

As mentioned, all the policies include `OrgOptConfig` in their own org config. This is then passed to the [`IsEnabled`](https://github.com/ossf/allstar/blob/main/pkg/config/config.go#L107) helper function along with the repo name. IsEnabled could get the repo info from GitHub (probably needs the org name as well), then include that in the calculation.

An example of getting the repo info is here: https://github.com/ossf/allstar/blob/main/pkg/policies/branch/branch.go#L158-L161

five510:
@jeffmendoza 
Thank you for support! I completed my implementation with your advice. Could you please review this PR?

jeffmendoza:
Sounds good, I'll take a look.

opt-out.md
```diff
@@ -7,13 +7,20 @@ To determine if Allstar is configured in the opt-out strategy, there will be a r
...
 
-To opt-out, submit a PR to that `.allstar` repo, and add the name of your repository to the opt-out list. ex:
+To opt-out specific repository, submit a PR to that `.allstar` repo, and add the name of your repository to the opt-out list. ex:
 
     optConfig:
       optOutStrategy: true
       optOutRepos:
       - my-repo-name-here
 
+To opt-out all private/public repositories, submit a PR to that `.allstar` repo, and add `optOutPrivateRepos` or `optOutPublicRepos`. ex:
```
jeffmendoza: I'd like to keep this file focused on a single repository opt-out. Please revert this, and add this section to the end of the section in https://github.com/ossf/allstar/blob/main/README.md#enable-configuration

pkg/config/config.go
```diff
@@ -97,20 +103,34 @@ func fetchConfig(ctx context.Context, r repositories, owner, repo, path string,
...
+		*github.Response, error)
 	GetContents(context.Context, string, string, string,
 		*github.RepositoryContentGetOptions) (*github.RepositoryContent,
 		[]*github.RepositoryContent, *github.Response, error)
 }
 
 // IsEnabled determines if a repo is enabled by interpreting the provided
 // org-level and repo-level OptConfigs.
-func IsEnabled(o OrgOptConfig, r RepoOptConfig, repo string) bool {
+func IsEnabled(ctx context.Context, o OrgOptConfig, r RepoOptConfig, rep repositories, owner, repo string) (bool, error) {
```
jeffmendoza: It's not good to have a public function with a parameter of a private type. Also, all the functions/packages only pass around `github.Client` in code currently. Make `IsEnabled()` take a `github.Client` then simply call a private `isEnabled()` which takes a `repositories`, then change each policy to pass in the client, not client.Repositories. The tests should then call the private `isEnabled` with the `repositories` mock object.

pkg/config/config_test.go
```diff
@@ -206,8 +254,16 @@ func TestIsEnabled(t *testing.T) {
 	}
 	for _, test := range tests {
 		t.Run(test.Name, func(t *testing.T) {
-			if IsEnabled(test.Org, test.Repo, "thisrepo") != test.Expect {
-				t.Errorf("Unexpected results. Expected: %v", test.Expect)
+			get = func(context.Context, string, string) (*github.Repository,
+				*github.Response, error) {
+				b := test.Org.OptOutPrivateRepos
+				return &github.Repository{
+					Private: &b,
```
jeffmendoza: This should be a separate input for each test, the mock response from github. Then the org config as you have it, then the expected output.

pkg/policies/binary/binary.go
```diff
@@ -107,7 +107,10 @@ func (l *logger) Debug(desc string, args ...interface{}) {
 func (b Binary) Check(ctx context.Context, c *github.Client, owner,
 	repo string) (*policydef.Result, error) {
 	oc, rc := getConfig(ctx, c, owner, repo)
-	enabled := config.IsEnabled(oc.OptConfig, rc.OptConfig, repo)
+	enabled, err := config.IsEnabled(ctx, oc.OptConfig, rc.OptConfig, c.Repositories, owner, repo)
+	if err != nil {
+		return nil, err
+	}
```
jeffmendoza, line 113: Looks like `config.IsEnabled` will need to be mocked now for all policies, similar to how `configFetchConfig` is currently mocked. It probably should have been mocked to begin with, as the policy tests don't need to unit test `IsEnabled`. The policy tests should now have the return value of `IsEnabled` be an input instead of the org/repo optconfigs.

jeffmendoza (changes requested):
Looking great so far. Updating the policy tests may be a bit tedious. Let me know if you'd like some help and I can update those.

opt-out.md
five510: Thanks. Updated with 48d36c7d664c0d821e438e19a94a511db9f054c6 and 75eee18e91963f2727285dbd0ab32ee42c4503e0

pkg/config/config.go
five510: It makes sense. Updated with a57b1ebf04d4dd118fb3cf5b7fa83a44a544f455.

pkg/policies/binary/binary.go
five510: Thanks. Updated with c59738bb78b1e2b816fed5dc7ab8f96feaedf30d.

pkg/config/config_test.go
five510: Thanks. Added mock config with d86295d7a81c9f9bdf84963965e9777719afa404.

five510:
@jeffmendoza Could you please check it again? Thanks!

jeffmendoza (approved):

jeffmendoza:
@five510 This deployed yesterday and looks like it is working great. Thanks for your contribution!!

PR #55, opened 2021-09-28
Log cache size and repo count every enforce loop.

jeffmendoza:
This also uses a single cache instead of separate cache per installation. I
believe the query params for listing the repos of an installation are different
so these won't collide.

### DIFF
--- a/pkg/enforce/enforce.go
+++ b/pkg/enforce/enforce.go
@@ -47,6 +47,7 @@ func init() {
 // TBD: determine if this should remain exported, or if it will only be called
 // from EnforceJob.
 func EnforceAll(ctx context.Context, ghc *ghclients.GHClients) error {
+	var repoCount int
 	ac, err := ghc.Get(0)
 	if err != nil {
 		return err
@@ -66,6 +67,10 @@ func EnforceAll(ctx context.Context, ghc *ghclients.GHClients) error {
 		}
 		opt.Page = resp.NextPage
 	}
+	log.Info().
+		Str("area", "bot").
+		Int("count", len(insts)).
+		Msg("Enforcing policies on installations.")
 	for _, i := range insts {
 		ic, err := ghc.Get(*i.ID)
 		if err != nil {
@@ -100,6 +105,12 @@ func EnforceAll(ctx context.Context, ghc *ghclients.GHClients) error {
 			continue
 		}
 		err = nil
+		log.Info().
+			Str("area", "bot").
+			Int64("id", *i.ID).
+			Int("count", len(repos)).
+			Msg("Enforcing policies on repos of installation.")
+		repoCount = repoCount + len(repos)
 		for _, r := range repos {
 			enabled := config.IsBotEnabled(ctx, ic, *r.Owner.Login, *r.Name)
 			err = RunPolicies(ctx, ic, *r.Owner.Login, *r.Name, enabled)
@@ -114,6 +125,11 @@ func EnforceAll(ctx context.Context, ghc *ghclients.GHClients) error {
 			continue
 		}
 	}
+	ghc.LogCacheSize()
+	log.Info().
+		Str("area", "bot").
+		Int("count", repoCount).
+		Msg("EnforceAll complete.")
 	return nil
 }
 
--- a/pkg/ghclients/ghclients.go
+++ b/pkg/ghclients/ghclients.go
@@ -24,7 +24,6 @@ import (
 	"github.com/google/go-github/v39/github"
 	"github.com/gregjones/httpcache"
 	"github.com/ossf/allstar/pkg/config/operator"
-	"github.com/rs/zerolog/log"
 	"gocloud.dev/runtimevar"
 	_ "gocloud.dev/runtimevar/gcpsecretmanager"
 )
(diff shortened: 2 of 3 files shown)


pkg/enforce/enforce.go
```diff
@@ -66,6 +67,10 @@ func EnforceAll(ctx context.Context, ghc *ghclients.GHClients) error {
 		}
 		opt.Page = resp.NextPage
 	}
+	log.Info().
+		Str("area", "bot").
+		Int("count", len(insts)).
```
inferno-chromium, line 72: should we use more prescriptive names e.g. installationCount, 
similar below one, like repoCount ?
and something specific for id, like XYZID

inferno-chromium (dismissed):

pkg/enforce/enforce.go
jeffmendoza: I feel like it would be too verbose to have a unique name for each tag-meaning across all logs, but maybe not, not sure.
I've been mostly querying based on the message name, then the tag name is essentially scoped to that message.

inferno-chromium: fine as-is, was a nit suggestion. those names are nice if used as stackdriver metrics. i think here it is just a log message, so fine either way.

pkg/ghclients/memorycache.go
```diff
@@ -0,0 +1,61 @@
...
+package ghclients
+
+import "sync"
+
+// memoryCache is an implemtation of httpcache.Cache that stores responses in
+// an in-memory map.  It is a copy of httpcache.MemoryCache but exposes Items
+// to allow for calculating and logging the cache size.
+type memoryCache struct {
+	mu    sync.RWMutex
+	Items map[string][]byte
```
azeemshaikh38: Nit: `Items` is not exported outside the package. Can be lowercase, i.e `items`

pkg/ghclients/ghclients.go
```diff
@@ -80,13 +83,24 @@ func (g *GHClients) Get(i int64) (*github.Client, error) {
...
 		Transport:           tr,
-		Cache:               httpcache.NewMemoryCache(),
+		Cache:               g.cache,
 		MarkCachedResponses: true,
 	}
 	g.clients[i] = github.NewClient(&http.Client{Transport: ctr})
 	return g.clients[i], nil
 }
 
+func (g *GHClients) LogCacheSize() {
```
azeemshaikh38, line 92: Nit: define a `getCacheSize()` fn on `memoryCache` struct and simply call that here.

pkg/ghclients/memorycache.go
```diff
@@ -0,0 +1,61 @@
...
+// Delete removes key from the cache
+func (c *memoryCache) Delete(key string) {
+	c.mu.Lock()
+	delete(c.Items, key)
+	c.mu.Unlock()
+}
+
+// newMemoryCache returns a new memoryCache that will store items in an
+// in-memory map
+func newMemoryCache() *memoryCache {
```
azeemshaikh38, line 73: Nit: update fn signature to return the interface name instead of implementation struct name: `func newMemoryCache() httpcache.Cache`

pkg/ghclients/ghclients.go
```diff
@@ -80,13 +83,24 @@ func (g *GHClients) Get(i int64) (*github.Client, error) {
 	}
 	ctr := &httpcache.Transport{
 		Transport:           tr,
-		Cache:               httpcache.NewMemoryCache(),
+		Cache:               g.cache,
```
azeemshaikh38, line 85: Maybe I'm misunderstanding the usage here, but it somewhat worries me that we are sharing a memory cache across installation IDs. High possibility of OOM.

pkg/ghclients/ghclients.go
```diff
@@ -45,6 +46,7 @@ type GHClients struct {
 	clients map[int64]*github.Client
 	tr      http.RoundTripper
 	key     []byte
+	cache   *memoryCache
```
azeemshaikh38, line 48: Nit: use `httpcache.Cache` instead of `*memoryCache`

azeemshaikh38 (dismissed):

pkg/ghclients/memorycache.go
jeffmendoza: It is accessed outside the struct.

pkg/ghclients/ghclients.go
jeffmendoza: Will do. Then Items can be private.

pkg/ghclients/memorycache.go
jeffmendoza: I want to keep the literal type in GHClients so I can access Items, or call the new log fn.

pkg/ghclients/ghclients.go
jeffmendoza: Switching from multiple maps to a single map here. Still using the same amount of memory as before. I'm worried out OOM to, which is why I am trying to log the size.

pkg/ghclients/ghclients.go
jeffmendoza: As mentioned, I need to have the literal type to either access Items or a log fn.

pkg/ghclients/memorycache.go
azeemshaikh38: SGTM. FWIW, you can use Golang type assertion to achieve the same - `ghcache.(*memoryCache)`

pkg/ghclients/ghclients.go
jeffmendoza: Done, PTAL.

azeemshaikh38: LGTM

azeemshaikh38 (approved):

Issue #44, opened 2021-09-01
Allow custom label on raised issue


jdhaines:
Allstar always creates it's issue(s) with a gray label called `allstar`.  We use an organization-wide label associated with security / risk that we'd like it to use.  I've read through the code and config files and I don't see how to configure allstar to use a custom label, but it seems like something that would be possible based on the way the config works currently.

Thanks for the great tool!

jeffmendoza:
Thanks for the suggestion, right now it is only configurable at the operator level. I agree it would be good to have this in the org config.

https://github.com/ossf/allstar/blob/main/pkg/config/operator/operator.go#L40

jeffmendoza:
Fixed by #68

PR #56, opened 2021-09-29
Feature:  Open issue with custom label and footer.

five510:
## Issue
https://github.com/ossf/allstar/issues/44

## Changes
- Open issue with custom label and footer.
  - Override by `IssueLabel` and `IssueFooter` in org/repo config.


pkg/policydef/policydef.go
```diff
@@ -68,4 +69,8 @@ type Policy interface {
 	// GetAction must return the configured action from the policy's config. No
 	// validation is needed by the policy, it will be done centrally.
 	GetAction(ctx context.Context, c *github.Client, owner, repo string) string
+
+	// GetOrgActionConfig must return the configuration for the action from org and repo config.
+	// The return value must follow configdef.OrgActionConfig.
+	GetOrgActionConfig(ctx context.Context, c *github.Client, owner, repo string) configdef.OrgActionConfig
```
five510, line 75: Looks policydef should not have dependency on config package. I made simply struct configdef to communicate `enforce <=> policy` and `enforce <=> issue`.
I'm not sure this approach is good or not honestly... Could you look into it 🙇 ?

pkg/config/config.go
```diff
@@ -56,6 +59,19 @@ type OrgOptConfig struct {
...
+// define the action configuration.
+type OrgActionConfig struct {
+	// IssueLabel : set to override GitHubIssueLabel in operator.go.
+	// GitHubIssueLabel is the label used to tag, search, and identify GitHub
+	// Issues created by the bot.
+	IssueLabel string `yaml:"issueLabel"`
+	
+	// IssueFooter : set to override GitHubIssueFooter in operator.go.
+	// IssueFooter is added to the end of GitHub issues.
+	IssueFooter string `yaml:"issueFooter"`
```
five510, line 72: The operator.go has default action config `GitHubIssueLabel` and `GitHubIssueFooter` so support both in this PR.

five510:
@jeffmendoza 
I made PR for https://github.com/ossf/allstar/issues/44. Could you please look into it 🙇 ?

jeffmendoza:
Awesome, thanks for working on this!
Do you think there is a good use-case where an org would want different labels / footers per-policy? I'm thinking it should be in the allstar.yaml only and apply to all the issues.

As for the footer. I'd like to keep the operator-set footer. For example, we would like to link back to this repo for the OSSF-run instance of Allstar. I think the ability for an org-owner to configure an additional footer is an excellent addition, and can be added to the issue text before or after the operator-set footer. Again, I don't see a use-case to use different footers per-policy, but let me know if you see one. Thanks!

five510:
@jeffmendoza 
Thanks for your checking!!

As for label, I prefer that label is configurable for org/repo level.
For example, my org I belong has 100+ repositories and each repositories has each different maintainer. The maintainer has different maintenance ways so it's preferable to configure by per org/repo.
As you mentioned, that configuration doesn't need to configure by per policies. I also agree label is only configured in allstar.yaml.


As for footer, I totally agree your opinion. The footer is only configured by the operator-set footer.
The reason I add customFooter is that there is use-case to add text into the issue. For example, the use case is to add the urls of internal documents and to mention someones.
But additional texts should not be added in the footer, I thought we will prepare new additionalText parameter and this issue shold be handled by other PRs.

As conclusion, I thought ActionConfig should be as follows

```
issueLabel
operator.go(default) < allstar.yml in org repo.(if define) < allstar.yml in repo level.(if define and not DisableRepoOverride)

issueFooter
  -> not support in this PR
```

jeffmendoza:
@five510 This sounds great. Go ahead and remove the label code from the individual policies and then we can get this in.

`additionalText` sounds good for the followup PR.

five510:
I just made other PR https://github.com/ossf/allstar/pull/68 for this issue because this PR includes unnecessary commits.
I will close this PR.

PR #68, opened 2021-10-10
feat: Custom label.

five510:
## Issue
https://github.com/ossf/allstar/issues/44

## Changes
- Open issue with custom label.
  - Override by `IssueLabel` by org/repo allstar.yml.


five510:
@jeffmendoza 
Could you check it 🙇 ?

jeffmendoza (changes requested):
Everything looks great. I think since the ActionConfig was removed from all the individual policy configs, now configdef.ActionConfig can be removed and config.ActionConfig can be used everywhere.

Do you see anywhere that just using config.ActionConfig will cause problems?

jeffmendoza:
@five510 I was just going to delete configdef then run a configdef -> config find-replace, but ended up changing a bit more.

- I removed ActionConfig. The actions are top-level concepts in Allstar, and don't need to be in sub-parts of config. The OptConfigs are used in all the policies, but this isn't. I usually don't like the idea of passing around structs as parameter lists for functions, preferring to list out all the params if needed in each function.
- I made the issue package call config directly, no need for this to be in the ensure package.
- I moved the label-calculating logic to the issue package. This is specific to issues. Config package is good for helper functions that can be used by multiple policies.
Please take a look and let me know if it looks good. Thanks.

jeffmendoza (approved):

jeffmendoza:
@asraa Can you do a code review? Thanks.

asraa (approved):
Thanks! Looks great to me! I tried very hard to nit :P

Issue #59, opened 2021-09-29
Where will issues be submitted?


Hixie:
For the flutter org, we have lots of repos but (with a few small exceptions) we use flutter/flutter/issues for all of them. Where will allstar try to submit issues if I enable that?

jeffmendoza:
Issues are submitted on the repo for which the policy violation is found.

Hixie:
Any chance we could have a configuration option to put issues into a different repo's issues database? Most of our repos have issues disabled so we wouldn't be able to get notifications in the current scheme.

jeffmendoza:
Yes, this sounds great. Currently #56 is adding an "ActionConfig" section to the org config for github labels. A configurable repo makes sense there as well.

jeffmendoza:
Custom label in #68, issues repo in #80

PR #80, opened 2021-11-15
Issues repo

### LINKED ISSUE #58: Preview mode
I administer an org with many, many repositories with varying levels of security (and varying levels of importance). I would love to enable allstar globally, but I am concerned that it will file hundreds of issues on various repos that may be secret, archived, unimportant, or some combination of the above, as well as filing issues that are known to be bogus (for example some of our repos are intentionally for storing binary files so the whole "don't store binary files" thing won't work in those repos).

What I would like is a way to get a preview of what allstar is going to complain about ahead of me doing something with public visibility.

jeffmendoza:
Add the `issueRepo` setting at the org level to direct all issue creation to a single repo for the whole org.

See https://github.com/jeffmendoza/.jtp/issues for an example.

Fixes #58 #59

### DIFF
--- a/README.md
+++ b/README.md
@@ -172,8 +172,8 @@ Two settings are available to configure the issue action:
   will override the default `allstar` label used by Allstar to identify its
   issues.
 
-- `issueRepo` is available at the orgainaztion level. Setting it will force all
-  issues created in the organization to be created in the repo specfied.
+- `issueRepo` is available at the organization level. Setting it will force all
+  issues created in the organization to be created in the repository specified.
 
 ## **Policies**
 
--- a/README.md
+++ b/README.md
@@ -164,6 +164,17 @@ future.
 - `email`: Allstar would send an email to the repository administrator(s).
 - `rpc`: Allstar would send an rpc to some organization-specific system.
 
+### **Action configuration**
+
+Two settings are available to configure the issue action:
+
+- `issueLabel` is available at the organization and repository level. Setting it
+  will override the default `allstar` label used by Allstar to identify its
+  issues.
+
+- `issueRepo` is available at the orgainaztion level. Setting it will force all
+  issues created in the organization to be created in the repo specfied.
+
 ## **Policies**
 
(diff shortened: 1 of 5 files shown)


README.md
```diff
@@ -164,6 +164,17 @@ future.
...
 
+### **Action configuration**
+
+Two settings are available to configure the issue action:
+
+- `issueLabel` is available at the organization and repository level. Setting it
+  will override the default `allstar` label used by Allstar to identify its
+  issues.
+
+- `issueRepo` is available at the orgainaztion level. Setting it will force all
```
laurentsimon: orgainaztion -> orginization

laurentsimon (dismissed):

README.md
jeffmendoza: Thanks ran spell check, PTAL.

azeemshaikh38 (approved):

inferno-chromium:
@jeffmendoza - minor fyi, i think you can merge the pull request here, instead of committing via local. that way you will see the PR correlation with your commit. e.g. https://github.com/ossf/allstar/commit/92486a876fa2d4bc2d943cb43082964d3495abf8 , you won't see that this PR is associated with your change.

README.md
```diff
@@ -164,6 +164,17 @@ future.
...
+### **Action configuration**
+
+Two settings are available to configure the issue action:
+
+- `issueLabel` is available at the organization and repository level. Setting it
+  will override the default `allstar` label used by Allstar to identify its
+  issues.
+
+- `issueRepo` is available at the organization level. Setting it will force all
+  issues created in the organization to be created in the repository specified.
```
Hixie, line 176: i've been looking everywhere but can't figure out what the syntax is for this setting -- is it just the name of the repo (e.g. "allstar"), a URL to the issues database, something else?

Hixie: (i see the code answers the question below but it'd be nice if the README was clearer)

PR #90, opened 2022-01-06
Create scorecards-analysis.yml

jeffmendoza:


laurentsimon (approved):
Thanks!

laurentsimon:
I'm curious: did you crate the PAT already or not? I'm trying to gauge whether the documentation is not explicit enough about crating it. cc @olivekl

laurentsimon:
Looks like you did not, the runs are failing

jeffmendoza:
I see now:
```
          # Read-only PAT token. To create it,
          # follow the steps in https://github.com/ossf/scorecard-action#pat-token-creation.
          repo_token: ${{ secrets.SCORECARD_TOKEN }}
```
I assumed this was something set already in the action environment. Why is this needed? Usually actions can get all the access they need by setting permissions in the workflow file. They would only need a secret set for something _outside_ of GitHub.

laurentsimon:
this is the reason https://github.com/ossf/scorecard/issues/1097. We found that branch protection settings with graphQl needs special permissions to be retrieved. We're going to revive the old REST APIs that work with the default GitHub token in v5 release (when scorecard is run in an action). Let me know if you find another way

PR #99, opened 2022-01-20
Update ko used for deployment.

jeffmendoza:


cloudbuild.yaml
```diff
@@ -1,7 +1,7 @@
 steps:
 - name: golang
   entrypoint: go
-  args: ['get', 'github.com/google/ko@v0.8.3']
+  args: ['install', 'github.com/google/ko@v0.9.3']
```
laurentsimon, line 4: no pinning by hash?

laurentsimon (approved):

cloudbuild.yaml
jeffmendoza: Hm, normally go.sum stores the hash, but I don't think it can be used here, so maybe hash is required

jeffmendoza: Just looking it up. The go command will check the checksum database (https://go.dev/ref/mod#checksum-database) unless  `-insecure` is used.

jeffmendoza: More info: https://go.dev/blog/module-mirror-launch
> Even the author of a module can’t move their tags around or otherwise change the bits associated with a specific version from one day to the next without the change being detected.

PR #98, opened 2022-01-20
Update to github.com/ossf/scorecard/v4

justaugustus:
Fixes https://github.com/ossf/allstar/issues/70, fixes https://github.com/ossf/allstar/issues/71.

- go.mod: Update to github.com/ossf/scorecard/v4@v4.0.1
- binary: Update logger methods to use `checker.LogMessage`
- binary: Drop deprecated fields from `checker.CheckRequest`
- binary: Stub SARIF log methods to satisfy `checker.DetailLogger`
- binary: Add `repoLogger` for `githubrepo.CreateGithubRepoClient`
- binary: Fixup `repoClient` args

Signed-off-by: Stephen Augustus <foo@auggie.dev>

xref: https://github.com/ossf/scorecard/issues/1273, https://github.com/ossf/allstar/issues/21

PR #100, opened 2022-01-20
Rollback 98

jeffmendoza:
Unfortunately the scorecard code is failing with
> 2022/01/20 20:10:18 GitHub token env var is not set. Please read https://github.com/ossf/scorecard#authentication

Somehow it is not using the passed in authenticated github client and still looking for the token env var.
Sorry for not having good integration tests here.


azeemshaikh38 (approved):

justaugustus:
breadcrumbs: https://github.com/ossf/allstar/pull/98

PR #106, opened 2022-01-26
Setup staging environment for Allstar.

jeffmendoza:
Using cloudbuild.yaml, Allstar will deploy to staging first with
app-staging.yaml The appengine yamls will set enviornment variables to
differientate between staging and prod. The pkg/config/operator package will
pull in these settings from the environment variables. There will be a seperate
yaml, similar to cloudbuild.yaml for deploying to prod using
app-prod.yaml. This will be entered directly into Cloud Build, and triggered
manually.

### DIFF
--- a/pkg/config/operator/operator.go
+++ b/pkg/config/operator/operator.go
@@ -15,17 +15,25 @@
 // Package operator contains config to be set by the GitHub App operator
 package operator
 
-import "time"
+import (
+	"os"
+	"strconv"
+	"time"
+)
 
 // AppID should be set to the application ID of the created GitHub App. See:
 // https://docs.github.com/en/developers/apps/building-github-apps/authenticating-with-github-apps#authenticating-as-a-github-app
-const AppID = 119816
+const setAppID = 119816
+
+var AppID int64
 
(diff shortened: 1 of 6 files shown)


pkg/config/operator/operator.go
```diff
@@ -53,3 +61,27 @@ Issue created by Allstar. See https://github.com/ossf/allstar/ for more informat
...
+
+var osGetenv func(string) string
+
+func init() {
+	osGetenv = os.Getenv
+	setVars()
+}
+
+func setVars() {
+	appIDs := osGetenv("APP_ID")
```
azeemshaikh38, line 73: Consider using `os.LookupEnv` which can tell you whether this var was set at all.

azeemshaikh38 (approved):
This is great, thanks!

pkg/config/operator/operator.go
jeffmendoza: Cool, In this case I would take the same action if the var is empty or not set, so I'll leave it. Thanks for pointing that one out, It will come in handy in the future.

Issue #113, opened 2022-02-10
Alerting for repos missing dependabot config (`dependabot.yml`)


justaugustus:
Similar to the `SECURITY.md` community health check, for organizations that are interested in having dependabot automated updates enabled on their repositories, it'd be great to have a check that alerts when a `.github/dependabot.yml` config is not present within the repo.

Ideally, it'd be cool to have this baked into `https://github.com/organizations/<org-name>/settings/security_analysis`.
cc: @jhutchings1

jeffmendoza:
:+1: https://github.com/ossf/allstar#future-policies
Also want to support renovatebot and others.

justaugustus:
> 👍 https://github.com/ossf/allstar#future-policies Also want to support renovatebot and others.

_starts to roll sleeves up..._ https://github.com/ossf/allstar/pull/114

jhutchings1:
> Ideally, it'd be cool to have this baked into `https://github.com/organizations/<org-name>/settings/security_analysis`.

We're looking at making this process simpler in the near future. @erinhav FYI on ☝🏻

ArisBee:
There is a way to do something like that by configuring the Scorecard check to only score on non-compliant dependency update tool, for instance in `scorecard.yaml` try:

```yaml
optConfig:
  optOutStrategy: true
action: log 
checks:
  - Dependency-Update-Tool
threshold: 5 
```

Jung47:
🙏🙏🙏

Issue #28, opened 2021-08-11
Feature: require GitHub Actions to be present


hazcod:
Would be great if allstar could verify whether certain GitHub Actions exist and a minimum version.
Often GitHub Actions contain a set of security screenings to block the PR.

jeffmendoza:
Yes, this is needed. Being able to generically configure an expected action/workflow to be present seems very useful. I believe defining how the config would work and the behavior of the policy will be tricky, help is needed here. For example, you can have multiple workflows in a repo, based on different triggers, and they may block/allow different things. I assume we would want to ensure a specific action is a part of a workflow that blocks something specific (PR merge, etc.).

justaugustus:
This should be implemented now.
See https://github.com/ossf/allstar?tab=readme-ov-file#github-actions.

Issue #22, opened 2021-07-27
Convert Binary Artifacts policy to be a  generic scorecard policy


jeffmendoza:
Should have multiple checks enabled (3-5 to start). Default config will be to run all enabled checks, but can override in config to only run certain ones. Should have a "success value" to decide what score (0-10) is passing.

jeffmendoza:
Do after #21

jeffmendoza:
I recently added some improved issues text to the Binary Artifacts policy. We will need to figure out a balance of keeping user-actionable text in the issues, and also opening up more/all of the Scorecard checks to Allstar users in a scaleable way.

jeffmendoza:
obsolete

PR #114, opened 2022-02-10
[WIP] policies: Add check for scorecard policies

justaugustus:
~Implements the scorecard [`Dependency Update Tool`](https://github.com/ossf/scorecard/blob/main/docs/checks.md#dependency-update-tool) check as an Allstar policy.~

~Fixes/ https://github.com/ossf/allstar/issues/113~

Signed-off-by: Stephen Augustus <foo@auggie.dev>


justaugustus:
Heh, I copied the SECURITY.md policy and assumed this was a little simpler than it was (roughly a graphql query for some URL).

This actually more closely mimics the logic for branch protection.
We'll likely want to:
- get a repo
- get if automated alerts enabled
- get file `.github/dependabot.yml`

jeffmendoza:
Looking good so far, let me know if you want an early review, or have any questions about anything.

justaugustus:
Turns out scorecard already has a policy for this: https://github.com/ossf/scorecard/blob/main/docs/checks.md#dependency-update-tool
So this is roughly copy/paste of the Binary Artifacts policy logic and now ready for review 🤞🏾 :)

jeffmendoza:
> So this is roughly copy/paste of the Binary Artifacts policy logic

Oh ok, so I actually have a couple of issues with this unfortunately.

For Scorecard integration, the initial thought was #22 and that policy would list the names of the checks. A main driver is that we didn't want to be downloading the repo tarball multiple times in Allstar. Also, to share code.

However, with the scorecard action that launched, it looks like a better path is #28. This will essentially enforce all checks in scorecard.

I do still think we can and should have a dependabot/renovatebot policy in Allstar if we gain anything by having access to the GitHub API, which scorecard does not get.

justaugustus:
FYI scorecard maintainers (@inferno-chromium @laurentsimon @naveensrinivasan @azeemshaikh38) --

I'm working on a refactor of the `scorecard` cmd to more easily allow consumers access to the same options that the CLI tool has. Once it's tighter, I'll lift it in the scorecard repo.

Opting to work in a temp directory in this PR because:
- I'm not positive of the final locations for various things
- It's cumbersome to do the changes, push to https://github.com/ossf/scorecard/pull/1645, then grab the new SHA here

azeemshaikh38:
> FYI scorecard maintainers (@inferno-chromium @laurentsimon @naveensrinivasan @azeemshaikh38) --
> 
> I'm working on a refactor of the `scorecard` cmd to more easily allow consumers access to the same options that the CLI tool has. Once it's tighter, I'll lift it in the scorecard repo.
> 
> Opting to work in a temp directory in this PR because:
> 
> * I'm not positive of the final locations for various things
> * It's cumbersome to do the changes, push to [✨ [WIP] Miscellaneous refactors to ease downstream consumption scorecard#1645](https://github.com/ossf/scorecard/pull/1645), then grab the new SHA here

I'm not sure I fully understand what this PR in AllStar is trying to achieve. Could you elaborate on:
- what's the goal in this PR
- how does it tie in with #28 which seems is the long term plan for AllStar

jeffmendoza:
Taking over in #240

PR #118, opened 2022-02-12
Update all actions to use commit hashes.

jauderho:
This updates the workflows to use only commit hashes. There is no other change and Dependabot should pick up any changes as needed going forward.

Signed-off-by: Jauder Ho <jauderho@users.noreply.github.com>


jeffmendoza:
Thanks for this!

jeffmendoza (approved):

PR #101, opened 2022-01-20
Update to github.com/ossf/scorecard/v4 (take two)

justaugustus:
- Revert "Rollback 98"
- go.mod: Update scorecard to v4.0.2-0.20220216001345-ba503c3bee01
- binary: Update arguments for `InitRepo`
- binary: Re-use existing `http.RoundTripper` for `clients.RepoClient`
- binary: Re-use scorecard `DetailLogger` implementation

Signed-off-by: Stephen Augustus <foo@auggie.dev>

---

From @jeffmendoza in https://github.com/ossf/allstar/pull/100#issue-827903307:

> Unfortunately the scorecard code is failing with
> 
> > 2022/01/20 20:10:18 GitHub token env var is not set. Please read https://github.com/ossf/scorecard#authentication
> 
> Somehow it is not using the passed in authenticated github client and still looking for the token env var. Sorry for not having good integration tests here.

This reverts https://github.com/ossf/allstar/pull/100 and will reapply https://github.com/ossf/allstar/pull/98, fixing the above comment.
Tracks https://github.com/ossf/scorecard/issues/1273, https://github.com/ossf/scorecard/pull/1502, and https://github.com/ossf/scorecard/pull/1516.
Re-closes https://github.com/ossf/allstar/issues/70.

PR #119, opened 2022-02-15
cmd: Use HTTP roundtripper from scorecard

justaugustus:
When scorecard moved to v2, support for older GitHub clients in
`CheckRequest` was dropped, which means any means of GitHub
authentication instantiated within allstar gets dropped as well.

Specifically, this affects the binary artifacts policy check.

Here, we opt to use scorecard's HTTP roundtripper in
`NewGHClients`, which should allow us to seed GitHub credentials
in the same way that scorecard does.

Unfortunately, this has the mild side-effect of temporarily
introducing zap as a logger for a small portion of the code (but
this will be removed again once we move to scorecard/v4).

Signed-off-by: Stephen Augustus <foo@auggie.dev>


justaugustus:
From https://twitter.com/stephenaugustus/status/1493660282886967297:

> I wrote some code yesterday and then realized (the second I closed my laptop), "Wow, this does not solve the problem at all!"

I think I need to expose some options in the scorecard GitHub client for these to play nicely together...

jeffmendoza:
I don't think we want to use the scorecard transport, we want to override it and provide our own. I just opened https://github.com/ossf/scorecard/pull/1644 which is what I think we need to then update the Allstar code.

justaugustus:
> I don't think we want to use the scorecard transport, we want to override it and provide our own. I just opened [ossf/scorecard#1644](https://github.com/ossf/scorecard/pull/1644) which is what I think we need to then update the Allstar code.

Agreed! Superseded by https://github.com/ossf/scorecard/pull/1644 and https://github.com/ossf/allstar/pull/101.

PR #131, opened 2022-03-04
Add Organization level custom footer

jeffmendoza:
Add to org level config a setting to allow specifying a custom footer text to
be included in all created issues in that org. This is useful to direct users
to org level settings, contacts, documentation, etc.


jeffmendoza:
cc @gauntface

azeemshaikh38 (approved):

Issue #30, opened 2021-08-12
Enforce certain status checks in branch protections


artis3n:
Accept an optional yaml list of status check names for the branch protection policy to enforce repos have enabled an organization's expected checks. The yaml list should accept objects containing at least the fields `name` and `required` (to enforce whether a particular status check should be marked as required).

I'm happy to take on this pull request myself if the maintainers approve of the issue.

artis3n:
Maybe also, in the object: an optional `msg` field that overwrites the message given to developers if that status check is found missing. If not present a general "missing X status check" message will be outputted.

jeffmendoza:
This sounds great. It would be good to propose the additional names and types of the members in [the config](https://pkg.go.dev/github.com/ossf/allstar@v0.0.0-20210812151424-aa093e731e2a/pkg/policies/branch#OrgConfig).

Why would `required` be needed. Why would you list a name of expected check, but it not be required?

The [GitHub data](https://pkg.go.dev/github.com/google/go-github/v38/github#RequiredStatusChecks) looks straight forward, just a list of strings.

artis3n:
> Why would `required` be needed. Why would you list a name of expected check, but it not be required?

We have a bunch of status checks we supply through GitHub Apps in our org to give feedback to folks as they work on PRs, but they don't necessarily need to succeed before code is merged. e.g. providing teams early security feedback but the violations only _have_ to be fixed later on, teams can still merge code with the failed status check.

I'm fine with putting this together without the `required` field but that's my use case.

artis3n:
Given the GitHub data it seems the status checks submitted must be considered required, which makes sense. So the new data going into the config would look like:

```yaml
status_checks:
  - mycheck
  - theothercheck
  - other
```

With no other metadata needed. Just the names

jeffmendoza:
Sounds great, lgtm

yorinasub17:
I had a need for this for our org's use case so I went ahead and took a stab at an implementation! https://github.com/ossf/allstar/pull/147

azeemshaikh38:
FYI, Scorecard `Branch-Protection` data already has the status checks. Not needed right now, but can consider pulling in this info from Scorecard directly at some point.

yorinasub17:
FYI #147 was just merged. Note that the final version changed the interface slightly to allow for specifying the GH app that should provide the context:

```yaml
requireStatusChecks:
  - context: "Analyze (go)"
    appID: 15368
  - context: "lint"
  - context: "build"
[...]

PR #147, opened 2022-03-22
Update branch protection policy to include required status checks

yorinasub17:
This implements https://github.com/ossf/allstar/issues/30 by introducing the following config settings to `branch_protection.yaml`:

- `requireUpToDateBranch`: Controls whether required status checks should be strict, as defined in [the GitHub documentation](https://docs.github.com/en/repositories/configuring-branches-and-merges-in-your-repository/defining-the-mergeability-of-pull-requests/about-protected-branches#require-status-checks-before-merging).
    - Note that (AFAIK) you can enable strict required status checks without any required checks, hence I set this as a separate flag.
    - I defaulted this to `false` for backward compatibility reasons.
- `requireStatusChecks`: List of status checks that should be required for merging.

When operating in fix mode, the following logic is implemented:
- If repo has no required status checks, and `requireUpToDateBranch = true`, enable strict required status checks.
- If repo has no required status checks, and `requireStatusChecks` is non-empty, set required status checks to the desired list.
- If repo has required status checks and strict mode is false but `requireUpToDateBranch = true`, enable strict mode.
- If repo has required status checks and `requireStatusChecks` is non-empty, merge in status check to the existing list, deduping at the end.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -6,7 +6,7 @@ require (
 	github.com/bradleyfalzon/ghinstallation/v2 v2.0.4
 	github.com/golang-jwt/jwt/v4 v4.1.0 // indirect
 	github.com/google/go-cmp v0.5.7
-	github.com/google/go-github/v39 v39.2.0
+	github.com/google/go-github/v43 v43.0.0
 	github.com/gregjones/httpcache v0.0.0-20190611155906-901d90724c79
 	github.com/onsi/ginkgo v1.16.5 // indirect
 	github.com/ossf/scorecard/v4 v4.0.2-0.20220216001345-ba503c3bee01
--- a/go.mod
+++ b/go.mod
@@ -8,7 +8,6 @@ require (
 	github.com/google/go-cmp v0.5.7
 	github.com/google/go-github/v43 v43.0.0
 	github.com/gregjones/httpcache v0.0.0-20190611155906-901d90724c79
-	github.com/onsi/ginkgo v1.16.5 // indirect
 	github.com/ossf/scorecard/v4 v4.0.2-0.20220216001345-ba503c3bee01
 	github.com/rs/zerolog v1.26.1
 	github.com/shurcooL/githubv4 v0.0.0-20210725200734-83ba7b4c9228
--- a/pkg/policies/branch/branch.go
+++ b/pkg/policies/branch/branch.go
@@ -122,6 +122,11 @@ type StatusCheck struct {
 	AppID *int64 `yaml:"appID"`
 }
 
+type statusCheckHash struct {
+	context string
+	appID   int64
+}
+
 type mergedConfig struct {
 	Action                string
 	EnforceDefault        bool
@@ -319,11 +324,13 @@ func check(ctx context.Context, rep repositories, c *github.Client, owner,
 				}
 				lt := makeSCLookupTable(rsc.Checks)
 				for _, c := range mc.RequireStatusChecks {
-					if _, ok := lt[c]; !ok {
-						appIDTxt := "(any app)"
-						if c.AppID != nil {
-							appIDTxt = fmt.Sprintf("(AppID: %v)", *c.AppID)
-						}
+					appIDTxt := "(any app)"
+					sch := statusCheckHash{context: c.Context}
+					if c.AppID != nil {
+						sch.appID = *c.AppID
+						appIDTxt = fmt.Sprintf("(AppID: %v)", *c.AppID)
+					}
+					if _, ok := lt[sch]; !ok {
 						text = text +
 							fmt.Sprintf("Status check %s %s not found for branch %v\n",
 								c.Context, appIDTxt, b)
@@ -504,7 +511,11 @@ func fix(ctx context.Context, rep repositories, c *github.Client,
 				lt := makeSCLookupTable(pr.RequiredStatusChecks.Checks)
 				for _, c := range mc.RequireStatusChecks {
 					// Only mark for update if there are status checks required, but not already set.
-					if _, ok := lt[c]; !ok {
+					sch := statusCheckHash{context: c.Context}
+					if c.AppID != nil {
+						sch.appID = *c.AppID
+					}
+					if _, ok := lt[sch]; !ok {
 						ac = append(ac, &github.RequiredStatusCheck{Context: c.Context, AppID: c.AppID})
 						update = true
 					}
@@ -615,14 +626,14 @@ func mergeConfig(oc *OrgConfig, rc *RepoConfig, repo string) *mergedConfig {
 	return mc
 }
 
-func makeSCLookupTable(prrsc []*github.RequiredStatusCheck) map[StatusCheck]struct{} {
-	lt := make(map[StatusCheck]struct{}, len(prrsc))
+func makeSCLookupTable(prrsc []*github.RequiredStatusCheck) map[statusCheckHash]struct{} {
+	lt := make(map[statusCheckHash]struct{}, len(prrsc))
 	for _, c := range prrsc {
 		// each check can match the context with and without appID
-		sc := StatusCheck{Context: c.Context, AppID: c.AppID}
+		sc := statusCheckHash{context: c.Context}
 		lt[sc] = struct{}{}
 		if c.AppID != nil {
-			lt[StatusCheck{Context: c.Context}] = struct{}{}
+			lt[statusCheckHash{context: c.Context, appID: *c.AppID}] = struct{}{}
 		}
 	}
 	return lt
--- a/pkg/policies/branch/branch.go
+++ b/pkg/policies/branch/branch.go
@@ -489,6 +489,8 @@ func fix(ctx context.Context, rep repositories, c *github.Client,
 						update = true
 					}
 				}
+				// Clear out Contexts, since API populates both, but updates require only one.
+				pr.RequiredStatusChecks.Contexts = nil
 				pr.RequiredStatusChecks.Checks = make([]*github.RequiredStatusCheck, 0)
 				for _, check := range allContexts {
 					pr.RequiredStatusChecks.Checks = append(pr.RequiredStatusChecks.Checks, check)
--- a/pkg/policies/branch/branch.go
+++ b/pkg/policies/branch/branch.go
@@ -101,8 +101,7 @@ type RepoConfig struct {
 	// RequireUpToDateBranch overrides the same setting in org-level, only if present.
 	RequireUpToDateBranch *bool `yaml:"requireUpToDateBranch"`
 
-	// RequireStatusChecks is a list of status checks (by name) that are required in
-	// order to merge into the protected branch.
+	// RequireStatusChecks overrides the same setting in org-level, only if present.
 	RequireStatusChecks []string `yaml:"statusChecks"`
 }
 
@@ -557,7 +556,6 @@ func mergeConfig(oc *OrgConfig, rc *RepoConfig, repo string) *mergedConfig {
 		RequireStatusChecks:   oc.RequireStatusChecks,
 	}
 	mc.EnforceBranches = append(mc.EnforceBranches, rc.EnforceBranches...)
-	mc.RequireStatusChecks = append(mc.RequireStatusChecks, rc.RequireStatusChecks...)
 
 	if !oc.OptConfig.DisableRepoOverride {
 		if rc.Action != nil {
@@ -580,6 +578,9 @@ func mergeConfig(oc *OrgConfig, rc *RepoConfig, repo string) *mergedConfig {
 		}
 		if rc.RequireUpToDateBranch != nil {
 			mc.RequireUpToDateBranch = *rc.RequireUpToDateBranch
+		}
+		if rc.RequireStatusChecks != nil {
+			mc.RequireStatusChecks = rc.RequireStatusChecks
 		}
 	}
 	return mc
--- a/pkg/policies/branch/branch.go
+++ b/pkg/policies/branch/branch.go
@@ -97,11 +97,15 @@ type RepoConfig struct {
 	// BlockForce overrides the same setting in org-level, only if present.
 	BlockForce *bool `yaml:"blockForce"`
 
-	// RequireUpToDateBranch overrides the same setting in org-level, only if present.
+	// RequireUpToDateBranch overrides the same setting in org-level, only if
+	// present.
 	RequireUpToDateBranch *bool `yaml:"requireUpToDateBranch"`
 
-	// RequireStatusChecks overrides the same setting in org-level, only if present.
-	RequireStatusChecks []string `yaml:"statusChecks"`
+	// RequireStatusChecks overrides the same setting in org-level, only if
+	// present. Omitting will lead to taking the org-level config as is, but
+	// specifying an empty list (`requireStatusChecks: []`) will override the
+	// setting to be empty.
+	RequireStatusChecks []string `yaml:"requireStatusChecks"`
 }
 
 type mergedConfig struct {
--- a/pkg/policies/branch/branch.go
+++ b/pkg/policies/branch/branch.go
@@ -63,7 +63,7 @@ type OrgConfig struct {
 	BlockForce bool `yaml:"blockForce"`
 
 	// RequireUpToDateBranch : set to true to require that branches must be up
-	// to date before merging. Default false.
+	// to date before merging. Only used if RequireStatusChecks is set. Default true.
 	RequireUpToDateBranch bool `yaml:"requireUpToDateBranch"`
 
 	// RequireStatusChecks is a list of status checks (by name) that are required in
@@ -98,8 +98,7 @@ type RepoConfig struct {
 	// BlockForce overrides the same setting in org-level, only if present.
 	BlockForce *bool `yaml:"blockForce"`
 
-	// RequireUpToDateBranch : set to true to require that branches must be up
-	// to date before merging. Default false.
+	// RequireUpToDateBranch overrides the same setting in org-level, only if present.
 	RequireUpToDateBranch *bool `yaml:"requireUpToDateBranch"`
 
 	// RequireStatusChecks is a list of status checks (by name) that are required in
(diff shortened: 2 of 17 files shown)


pkg/policies/branch/branch.go
```diff
@@ -60,6 +61,14 @@ type OrgConfig struct {
 
 	// BlockForce : set to true to block force pushes, default true.
 	BlockForce bool `yaml:"blockForce"`
+
+	// RequireUpToDateBranch : set to true to require that branches must be up
+	// to date before merging. Default false.
+	RequireUpToDateBranch bool `yaml:"requireUpToDateBranch"`
```
jeffmendoza, line 67: I'm thinking that this should default to true, but should only take effect if the RequireStatusChecks array length is non-zero.

jeffmendoza: The GitHub UI text is: "This setting will not take effect unless at least one status check is enabled (see below)."

pkg/policies/branch/branch.go
```diff
@@ -88,23 +97,35 @@ type RepoConfig struct {
 
 	// BlockForce overrides the same setting in org-level, only if present.
 	BlockForce *bool `yaml:"blockForce"`
+
+	// RequireUpToDateBranch : set to true to require that branches must be up
+	// to date before merging. Default false.
+	RequireUpToDateBranch *bool `yaml:"requireUpToDateBranch"`
```
jeffmendoza, line 104: Update the comment to indicate how the repo-config is merged with the org-config. Ex:
`RequireUpToDateBranch overrides the same setting in org-level, only if present.`

pkg/policies/branch/branch.go
```diff
@@ -88,23 +97,35 @@ type RepoConfig struct {
...
+
+	// RequireStatusChecks is a list of status checks (by name) that are required in
+	// order to merge into the protected branch.
+	RequireStatusChecks []string `yaml:"statusChecks"`
```
jeffmendoza: Update the comment to indicate how the repo-config is merged with the org-config. Ex:
```
RequireStatusChecks adds more check names to the org-level list. Does not
override. Always allowed irrespective of DisableRepoOverride setting.
```
Secondly, is this the behavior that we want? Is there a situation where a repo might want to remove/replace the org-level list?

pkg/policies/branch/branch.go
```diff
@@ -456,15 +550,18 @@ func getConfig(ctx context.Context, c *github.Client, owner, repo string) (*OrgC
...
+		EnforceBranches:       oc.EnforceBranches[repo],
+		RequireApproval:       oc.RequireApproval,
+		ApprovalCount:         oc.ApprovalCount,
+		DismissStale:          oc.DismissStale,
+		BlockForce:            oc.BlockForce,
+		RequireUpToDateBranch: oc.RequireUpToDateBranch,
+		RequireStatusChecks:   oc.RequireStatusChecks,
 	}
 	mc.EnforceBranches = append(mc.EnforceBranches, rc.EnforceBranches...)
+	mc.RequireStatusChecks = append(mc.RequireStatusChecks, rc.RequireStatusChecks...)
```
jeffmendoza: See comment on definition, do we simply want to append here, or allow replacement (only with repo override allowed)

pkg/policies/branch/branch.go
```diff
@@ -267,6 +288,42 @@ func check(ctx context.Context, rep repositories, c *github.Client, owner,
...
+			d.RequireUpToDateBranch = false
+			if mc.RequireUpToDateBranch {
+				text = text +
+					fmt.Sprintf("Require up to date branch not configured for branch %v\n",
+						b)
+				pass = false
+			}
+			if len(mc.RequireStatusChecks) > 0 {
+				text = text +
+					fmt.Sprintf("No status checks required for branch %v\n", b)
```
jeffmendoza: I might re-word this to: "Status checks required by policy, but none found."

pkg/policies/branch/branch.go
```diff
@@ -267,6 +288,42 @@ func check(ctx context.Context, rep repositories, c *github.Client, owner,
...
+				if _, ok := c[check]; !ok {
+					text = text +
+						fmt.Sprintf("Status check %v not required for branch %v\n",
+							check, b)
+					pass = false
+				}
+			}
+		} else {
+			d.RequireUpToDateBranch = false
+			if mc.RequireUpToDateBranch {
```
jeffmendoza: So, I'd think we remove this case, if we make the above change that this is only applicable if `len(mc.RequireStatusChecks) > 0`

pkg/policies/branch/branch.go
```diff
@@ -267,6 +288,42 @@ func check(ctx context.Context, rep repositories, c *github.Client, owner,
...
+			}
+			d.RequireStatusChecks = rsc.Contexts
+			c := make(map[string]struct{}, len(rsc.Contexts))
+			for _, ctx := range rsc.Contexts {
+				c[ctx] = struct{}{}
+			}
+			for _, check := range mc.RequireStatusChecks {
+				if _, ok := c[check]; !ok {
+					text = text +
+						fmt.Sprintf("Status check %v not required for branch %v\n",
```
jeffmendoza: Maybe: "Status check %v not found for branch %v\n"?

pkg/policies/branch/branch.go
```diff
@@ -267,6 +288,42 @@ func check(ctx context.Context, rep repositories, c *github.Client, owner,
 				d.BlockForce = false
 			}
 		}
+		rsc := p.GetRequiredStatusChecks()
+		if rsc != nil {
+			d.RequireUpToDateBranch = rsc.Strict
+			if mc.RequireUpToDateBranch && !rsc.Strict {
```
jeffmendoza: Would update this to `if mc.RequireUpToDateBranch && !rsc.Strict && len(mc.RequireStatusChecks) > 0`

pkg/policies/branch/branch.go
```diff
@@ -267,6 +288,42 @@ func check(ctx context.Context, rep repositories, c *github.Client, owner,
...
+				text = text +
+					fmt.Sprintf("Require up to date branch not configured for branch %v\n",
+						b)
+				pass = false
+			}
+			d.RequireStatusChecks = rsc.Contexts
+			c := make(map[string]struct{}, len(rsc.Contexts))
```
jeffmendoza: The [library docs](https://pkg.go.dev/github.com/google/go-github/v43/github#RequiredStatusChecks) say that `Contexts` is deprecated. Should probably use [`Checks`](https://pkg.go.dev/github.com/google/go-github/v43/github#RequiredStatusCheck).

Also, what do you think about using App ID instead? Could the check name have collisions? Is getting the right App ID too cumbersome for users?

pkg/policies/branch/branch.go
```diff
@@ -319,6 +376,14 @@ func fix(ctx context.Context, rep repositories, c *github.Client,
 					}
 					pr.RequiredPullRequestReviews = rq
 				}
+				if mc.RequireUpToDateBranch || len(mc.RequireStatusChecks) > 0 {
```
jeffmendoza: Change to `if len(mc.RequireStatusChecks) > 0`

pkg/policies/branch/branch.go
```diff
@@ -319,6 +376,14 @@ func fix(ctx context.Context, rep repositories, c *github.Client,
...
+					rsc := &github.RequiredStatusChecks{
+						Strict: mc.RequireUpToDateBranch,
+						// Can't be set to nil, so force non-nil value.
```
jeffmendoza: Wouldn't need this comment/append now with above change.

pkg/policies/branch/branch.go
```diff
@@ -397,6 +462,35 @@ func fix(ctx context.Context, rep repositories, c *github.Client,
 				update = true
 			}
 		}
+		if pr.RequiredStatusChecks == nil && (mc.RequireUpToDateBranch || len(mc.RequireStatusChecks) > 0) {
+			rsc := &github.RequiredStatusChecks{
```
jeffmendoza: Same as above.

pkg/policies/branch/branch.go
```diff
@@ -397,6 +462,35 @@ func fix(ctx context.Context, rep repositories, c *github.Client,
...
+			}
+			if len(mc.RequireStatusChecks) > 0 {
+				allContexts := make(map[string]struct{}, len(pr.RequiredStatusChecks.Contexts))
+				for _, ctx := range pr.RequiredStatusChecks.Contexts {
+					allContexts[ctx] = struct{}{}
+				}
+				for _, ctx := range mc.RequireStatusChecks {
+					allContexts[ctx] = struct{}{}
+				}
+				pr.RequiredStatusChecks.Contexts = make([]string, 0)
```
jeffmendoza: Again, should use Checks.Context.
https://docs.github.com/en/rest/reference/branches#update-status-check-protection
Looks like we can null out the AppID.

jeffmendoza (changes requested):
Awesome, looking very good, thanks for contributing!

I think we should definitely change the RequireUpToDateBranch to default to true but only have any effect if the user has one or more required status checks in Allstar config. This would mirror how GitHub works.

pkg/policies/branch/branch.go
yorinasub17: So I noticed this, but this project is currently using [`v39` of the `go-github` library](https://github.com/ossf/allstar/blob/main/go.mod#L9) which unfortunately doesn't expose `Checks` yet (`Checks` was introduced in [43.0.0](https://github.com/google/go-github/releases/tag/v43.0.0)). I was concerned about upgrading a fundamental library in a new project I never contributed to, which was why I went with the `Contexts` version.

With that said, if you think it is worth bumping the library in this PR, I'm happy to take a stab at it, including converting this to using `Checks` and requiring the list to be specified by App ID instead of Name.

pkg/policies/branch/branch.go
yorinasub17: Ah I totally missed that UI text. Thanks for pointing it out! I'll work on updating the logic to reflect your comment.

pkg/policies/branch/branch.go
jeffmendoza: > With that said, if you think it is worth bumping the library in this PR, I'm happy to take a stab at it, including converting this to using Checks

Yes that would be great!

> requiring the list to be specified by App ID instead of Name

Still up to you here, I'm not sure what is best, if the App ID would be better as not ambiguous, but would be more troublesome for the user to look up.

yorinasub17: Ok I managed to bump the go-github package version, and switched to using `Checks` instead of `Contexts` (a57c5c2d9b7c76ec7af91d9dd5375b39f3c090b1).

In regards to App ID though...

> Still up to you here, I'm not sure what is best, if the App ID would be better as not ambiguous, but would be more troublesome for the user to look up.

I looked into this, and I realized that we might actually want to expose both.

Context by itself is not sufficient, because multiple GitHub Apps could use the same context, and so it won't be obvious which one you are referring to. Worse, if you don't specify the GitHub App ID, then GitHub will simply pick the one that provided the context most recently - this might be exploitable by a user who runs a random github app that appends a check with the same context to the commit. Not 100% sure if that is possible since the docs don't specify the scope of the search. This is mostly an issue for the fix action.

OTOH, App ID by itself doesn't seem sufficient because a single App can provide multiple checks, so it won't be possible to limit the checks. E.g., all the checks running in GitHub actions would be provided by App ID `15368`, but our org wants to ignore one of the checks in our suite because of flaky testing results from cloud infrastructure testing.

Given all that, what are your thoughts about changing `requireStatusChecks` to be a list of structs where context is required, but app id is optional? E.g.:

```yaml
requireStatusChecks:
  - context: "Analyze (go)"
    appID: 15368
  - context: "lint"
  - context: "build"
    appID: 15368
```

pkg/policies/branch/branch.go
yorinasub17: Updated to only look at `RequireUpToDateBranch` if `RequireStatusChecks` is set: 754144bd3f9e24f3554793450990d44d4116b98f

pkg/policies/branch/branch.go
yorinasub17: Fixed in 754144bd3f9e24f3554793450990d44d4116b98f

pkg/policies/branch/branch.go
yorinasub17: Fixed in f902892e5f9bf4315541c3f48c04061c0648cd77

> Secondly, is this the behavior that we want? Is there a situation where a repo might want to remove/replace the org-level list?

This was a good point, and after thinking about it, I switched to a replacement instead of merging. If we do merging, then there is no way for the repo override to remove checks that may not be relevant to the repo (e.g., a go lint check on a nodejs project). At least with replacement, you can simulate a merge by repeating the list from the org.

pkg/policies/branch/branch.go
yorinasub17: Fixed in f902892e5f9bf4315541c3f48c04061c0648cd77

pkg/policies/branch/branch.go
yorinasub17: Updated in 754144bd3f9e24f3554793450990d44d4116b98f

pkg/policies/branch/branch.go
yorinasub17: Fixed in 754144bd3f9e24f3554793450990d44d4116b98f

pkg/policies/branch/branch.go
yorinasub17: Updated in 754144bd3f9e24f3554793450990d44d4116b98f

pkg/policies/branch/branch.go
yorinasub17: Updated in 754144bd3f9e24f3554793450990d44d4116b98f by having `len(mc.RequireStatusChecks) > 0` at the top level.

pkg/policies/branch/branch.go
yorinasub17: Fixed in 754144bd3f9e24f3554793450990d44d4116b98f

pkg/policies/branch/branch.go
yorinasub17: Yup! Update to `Checks` removed the need for this in a57c5c2d9b7c76ec7af91d9dd5375b39f3c090b1

pkg/policies/branch/branch.go
yorinasub17: Updated in https://github.com/ossf/allstar/commit/a57c5c2d9b7c76ec7af91d9dd5375b39f3c090b1

go.mod
```diff
@@ -6,9 +6,8 @@ require (
 	github.com/bradleyfalzon/ghinstallation/v2 v2.0.4
 	github.com/golang-jwt/jwt/v4 v4.1.0 // indirect
 	github.com/google/go-cmp v0.5.7
-	github.com/google/go-github/v39 v39.2.0
+	github.com/google/go-github/v43 v43.0.0
 	github.com/gregjones/httpcache v0.0.0-20190611155906-901d90724c79
-	github.com/onsi/ginkgo v1.16.5 // indirect
```
yorinasub17, line 11: FYI this was automatically removed when `go mod tidy` was run.

pkg/policies/branch/branch.go
```diff
@@ -485,6 +588,12 @@ func mergeConfig(oc *OrgConfig, rc *RepoConfig, repo string) *mergedConfig {
 		if rc.BlockForce != nil {
 			mc.BlockForce = *rc.BlockForce
 		}
+		if rc.RequireUpToDateBranch != nil {
+			mc.RequireUpToDateBranch = *rc.RequireUpToDateBranch
+		}
+		if rc.RequireStatusChecks != nil {
```
jeffmendoza, line 622:
```
if len(rc.RequireStatusChecks) > 0 {
```
For clarity because it is a slice and not a pointer.

pkg/policies/branch/branch.go
jeffmendoza: > this might be exploitable by a user who runs a random github app that appends a check with the same context to the commit

Yes, exactly.

> what are your thoughts about changing requireStatusChecks to be a list of structs where context is required, but app id is optional? 

This looks great! Lets go with that struct if you are up for it.

pkg/policies/branch/branch.go
yorinasub17: Actually, the `!= nil` is intentional to add support for the case of stripping all the required checks defined at the org level at the repo level. The `yaml` package explicitly sets the attribute to `nil` instead of `[]string{}` when it's omitted.

E.g., one could have the config:
```yaml
requireStatusChecks: []
```

See https://go.dev/play/p/7RsvjHP1BY6

Should I add a comment here to make that clear?

pkg/policies/branch/branch.go
yorinasub17: Cool! Will take a stab at implementing that!

pkg/policies/branch/branch.go
jeffmendoza: Ah, very cool! Yes, update the repoconfig comment on that to say that omitting leads to taking the orgconfig as is, but specifying an empty list will override the orgconfig to be empty. Thanks!

yorinasub17: Ok updated the comment: [c5b0001](https://github.com/ossf/allstar/pull/147/commits/c5b0001651646a1aeb95d710a90832266623e3f7)

pkg/policies/branch/branch.go
yorinasub17: Ok implemented the struct version: [fd13cad](https://github.com/ossf/allstar/pull/147/commits/fd13cad2b3ee6b6c258772a6ed2a8f545ddbb2cd)

~I tested this out on our org and it behaves as expected!~ ~Ah scratch that - it's not detecting the composite correctly.~ This bug is now fixed in c02b468c8b3183db109f675785887d51ffba2bb5

jeffmendoza (approved):

jeffmendoza:
Awesome, let's try out in staging.

jeffmendoza:
@yorinasub17 Staging is deployed: https://github.com/apps/allstar-staging
Any chance you can do a quick sanity check with that app instance? Thanks.
If good, I'll push to prod.

yorinasub17:
Just installed it on our test org (I can't install it into our main org for internal policy reasons) and looks like it ran successfully!

For the config:

```
optConfig:
  optOutStrategy: true
  disableRepoOverride: true
action: fix

enforceDefault: true

requireApproval: true
approvalCount: 1
dismissStale: true

blockForce: true

requireUpToDateBranch: false
requireStatusChecks:
  - context: lint
    appID: 15368
```

Result:

Screen Shot 2022-03-24 at 1 10 40 PM

jeffmendoza:
Great kicking off prod deployment now.

PR #148, opened 2022-03-22
Add ability to run checks only once instead of continuously

yorinasub17:
I'm not sure if this is in the best interest of the project, so feel free to close if this is veering too far.

When iterating in development, I found it useful to be able to run the `allstar` check only once through all the repos. This also helps with alternative deployment models, such as running the process on serverless AWS Lambda function on a periodic invoke schedule, where you want the task to be an ephemeral job instead of a continuous service.

This PR updates the CLI to accept a `-once` flag that can be used to invoke the enforce all routine once rather than a continuous loop.

### DIFF
--- a/cmd/allstar/main.go
+++ b/cmd/allstar/main.go
@@ -16,6 +16,7 @@ package main
 
 import (
 	"context"
+	"flag"
 	"net/http"
 	"os"
 	"os/signal"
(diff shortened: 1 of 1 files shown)


jeffmendoza (approved):

jeffmendoza:
Very cool, Thanks!

Issue #155, opened 2022-04-04
Bug in setting branch protection


jeffmendoza:
```
PUT https://api.github.com/repos/dart-lang/dart-services/branches/master/protection: 422 Invalid request.

No subschema in "anyOf" matched.
More than one subschema in "oneOf" matched.
Not all subschemas of "allOf" matched.
For 'anyOf/1', {"strict"=>true, "contexts"=>["cla/google", "analyze (stable)"], "checks"=>[{"context"=>"cla/google"}, {"context"=>"analyze (stable)"}]} is not a null. []
```
Looks like it is not copying over the existing required checks? Not sure if related to #147 or not. Digging in.

jeffmendoza:
Reproduced on my test repo:
```
PUT https://api.github.com/repos/jeffmendoza/test-settings/branches/main/protection: 422 Invalid request.                                                                                  
                                                                                                                                                                                           
No subschema in "anyOf" matched.                                                                                                                                                           
More than one subschema in "oneOf" matched.                                                                                                                                                
Not all subschemas of "allOf" matched.                                                                                                                                                     
For 'anyOf/1', {"strict"=>true, "contexts"=>["Inclusive Language"], "checks"=>[{"context"=>"Inclusive Language", "app_id"=>75645}]} is not a null. []                                      
```

PR #160, opened 2022-04-05
Update test for existing status checks.

jeffmendoza:
Test to catch #155 #156

### DIFF
--- a/pkg/policies/branch/branch_test.go
+++ b/pkg/policies/branch/branch_test.go
@@ -60,8 +60,6 @@ func (m mockRepos) UpdateBranchProtection(ctx context.Context, owner, repo,
 }
 
 func TestCheck(t *testing.T) {
-	one := 1
-	fal := false
 	tests := []struct {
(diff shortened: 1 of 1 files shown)


pkg/policies/branch/branch_test.go
```diff
@@ -651,6 +651,8 @@ func TestCheck(t *testing.T) {
 
 func TestFix(t *testing.T) {
 	flse := false
+	var appid int64
+	appid = 123
```
azeemshaikh38: Nit: use single line definition: `var appid int64 = 123`

azeemshaikh38 (dismissed):

pkg/policies/branch/branch_test.go
jeffmendoza: Ah cool! Looks like there are already some helper funcs to get pointers in the github library, switched to those.

azeemshaikh38 (approved):
Nice, didn't know this existed.

PR #168, opened 2022-04-12
Allow AWS Secrets Manager as a source for key secret

yorinasub17:
Due to company policy reasons, we've been running our own instance of `allstar`. However, we use AWS as our cloud, and the last thing that prevents us from using the upstream package vs our fork is the lack of support for reading the GitHub secrets from something other than GCP secret manager. This PR adds the driver support to use AWS secrets manager as an alternative source for the secret key.

Would be greatly appreciated if we can get this driver import in upstream so we don't have to maintain the fork for our deployments! (but understand if there is a desire to keep the binary light)


azeemshaikh38 (approved):
LGTM, thanks @yorinasub17! @jeffmendoza for a second review.

jeffmendoza:
Happy to add anything that gocloud.dev supports. Thanks for helping out other AWS folks!

yorinasub17:
Many thanks!

PR #169, opened 2022-04-12
Add Dangerous Workflow policy

jeffmendoza:
Adds new policy from Scorecard check. Includes scorecard package to store
client across policy runs.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -7,7 +7,7 @@ require (
 	github.com/google/go-cmp v0.5.7
 	github.com/google/go-github/v43 v43.0.0
 	github.com/gregjones/httpcache v0.0.0-20190611155906-901d90724c79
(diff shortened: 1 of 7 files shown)


jeffmendoza:
workflow.go has a lot of shared code with binary.go. I'm going to think more about a generic policy vs. these explicit ones. If we will want more explicit policies per scorecard check, I'll find a way to share the code properly. Otherwise, I'll work on a generic policy that can have custom settings and messages.

jeffmendoza:
I want to update the scorecards version then pull in https://github.com/ossf/scorecard/blob/eedd16d5be62a0dfdb48410a44d67d367263e027/checker/detail_logger_impl.go#L23 then re-open this PR.

jeffmendoza:
Updated scorecards to 4.1.0, but looks like the detaillogger constructor is not there yet, so leaving those there for now.

laurentsimon (approved):

PR #172, opened 2022-04-16
Disabled should not run anything

### LINKED ISSUE #171: Disabled checks shouldn't do anything
I am observing that if I opt out of a repo on a check, `allstar` still runs the check but in log mode. E.g., I have the new `Dangerous Workflow` check disabled, but I am seeing the error log `"unable to get tarball tarball not found"` for all my repos.

This is problematic for us for two reasons:
- We have a lot of repos, and many of them don't need to do any checks. However, each of these read-only checks eat into the runtime and GH API limits. It takes 15+ minutes right now to go through all our repos.
- It adds a lot of noise to the logs. The `unable to get tarball` log shows up for every repo, which is a lot for us. We also see the check logs for each of the areas for each repo, even for the ones that are globally opted out with the new opt-out-of-archived-repos setting. This isn't a huge deal if not for the fact that these logs can add up to the cost of our log aggregation system.

Note that I think this is only true for a subset of the checks. So far, I am observing this for the following checks:

- `SECURITY.md`
- `Dangerous Workflow`
- `Branch Protection`

The others appear to be doing the right thing by emitting only a `Disabled` log message.

---

If this seems reasonable, happy to submit a PR to go through the checks and make it do the same thing as the `Binary Artifacts` check.

yorinasub17:
Fixes https://github.com/ossf/allstar/issues/171

This PR updates all the policies to disable the check completely when it is not enabled via the config, like the `Binary Artifacts` check.

I see now that this was implemented this way because the other checks are relatively quick to run, but as mentioned on the issue, it is probably a good idea to bypass the checks if not needed to run to reduce API calls and avoid hitting rate limits.

### DIFF
--- a/pkg/config/operator/operator.go
+++ b/pkg/config/operator/operator.go
@@ -49,6 +49,15 @@ const RepoConfigDir = ".allstar"
 // AppConfigFile is the name of the expected file in org or repo level config.
 const AppConfigFile = "allstar.yaml"
 
+// DoNothingOnOptOut is a configuration flag indicating if allstar should do
+// nothing and skip the corresponding checks when a repository is opted out.
+// Can be configured with environment variable DO_NOTHING_ON_OPT_OUT, where
+// the value should be a string equivalent of a bool, as accepted by
+// strconv.ParseBool.
+const setDoNothingOnOptOut = false
+
+var DoNothingOnOptOut bool
+
 // GitHubIssueLabel is the label used to tag, search, and identify GitHub
 // Issues created by the bot.
 const GitHubIssueLabel = "allstar"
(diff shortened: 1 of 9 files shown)


jeffmendoza:
@yorinasub17 Thanks for taking a look at this. I'd like this to be operator configurable. Could you add an option to `pkg/config/operator` same as appid/secret?

yorinasub17:
Thanks for the review!

> Thanks for taking a look at this. I'd like this to be operator configurable. Could you add an option to pkg/config/operator same as appid/secret?

Make sense. I took a stab at implementing the operator option: let me know what you think! [81d1876](https://github.com/ossf/allstar/pull/172/commits/81d18761b109e0cd77d0ecef4b37494a3552159d)

jeffmendoza (approved):

jeffmendoza:
@yorinasub17 Thanks for this, looks great!

PR #192, opened 2022-05-26
Individual repo config in org config repo

### LINKED ISSUE #186: Proposal: Individual repo config in org config repo
## Problem

In our org, we would like to restrict access to the Allstar configurations as much as possible, which means that we want to set `disableRepoOverride = true` for all the policies. However, this leaves a problem where the settings for Allstar must be all or nothing - that is, we can only have one configuration for all the repos, or none at all.

This proves a challenge for the branch protection settings, where some repos have additional checks that we would like to enforce, and want to use Allstar to ensure those settings stay on. If we had repo level config, this can be handled by adding a `.allstar` folder and config in the underlying repos, but then this means that developers of the repo can turn Allstar checks off if they wanted to (e.g., set `action: log`), which we would like to avoid.

This is a request for a mechanism that allows configuring individual repo overrides in the org level `.allstar` repo.

## Solution

The proposed solution is to allow configuring individual repo overrides from the org `.allstar` repo by introducing each individual repo config as folders.

Consider the following proposed folder tree for a `.allstar` repo:

```
├── README.md
├── allstar.yaml
├── binary_artifacts.yaml
├── branch_protection.yaml
├── dangerous_workflow.yaml
├── outside.yaml
├── security.yaml
├── myapp
│   └── branch_protection.yaml
└── someotherapp
    ├── dangerous_workflow.yaml
    └── outside.yaml
```

In this tree, the global org level config is defined in the root `yaml` files as before. However, in this proposal, we also introduce individual repo folders, such as `myapp` and `someotherapp`, whose configurations will act as repo level configurations that can override the org level config.

In this example, `myapp/branch_protection.yaml` will be merged up with `binary_protection.yaml` to produce the final config.

Note that if `disableRepoOverride = false` and there exists a `.allstar/branch_protection.yaml` file in the `myapp` repo, then that will have precedence over `myapp/branch_protection.yaml`.

So the order of look up will be:

1. `myapp` repo  `.allstar/branch_protection.yaml`
2. org `.allstar` repo `myapp/branch_protection.yaml`
3. org `.allstar` repo `branch_protection.yaml`

---

If this solution seems reasonable, I'd be happy to work on a PR to implement this!

yorinasub17:
Fixes https://github.com/ossf/allstar/issues/186

This implements the proposed solution described in https://github.com/ossf/allstar/issues/186

### DIFF
--- a/README.md
+++ b/README.md
@@ -340,6 +340,25 @@ secondary location. To clarify, for `allstar.yaml`:
 This is also true for the org-level configuration files for the individual
 policies, as described below.
 
+### Repo policy configurations in the Org Repo
+
+Allstar will also look for repo-level policy configurations in the
+organization's `.allstar` repository, under the directory with the same name as
+the repository. This configuration is used regardless of whether "repo override"
+is disabled.
+
+For example, Allstar will lookup the policy configuration for a given repo
+`myapp` in the following order:
+
+| Repository | Path | Condition |
+| - | - | - |
+| `myapp` | `.allstar/branch_protection.yaml` | When "repo override" is allowed. |
+| `.allstar` | `myapp/branch_protection.yaml` | All times. |
+| `.allstar` | `branch_protection.yaml` | All times. |
+| `.github` | `allstar/myapp/branch_protection.yaml` | If `.allstar` repo does not exist. |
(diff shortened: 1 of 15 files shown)


jeffmendoza:
Big change! Everything looks great. I'll check it out in staging.

jeffmendoza (approved):

PR #193, opened 2022-05-26
Fix logic error with Binary Artifacts

jeffmendoza:


pkg/policies/binary/binary.go
```diff
@@ -182,7 +182,7 @@ func (b Binary) Check(ctx context.Context, c *github.Client, owner,
 	logs := convertAndFilterLogs(l.logs, mc)
 
 	// We assume every log is a finding and do filtering on the Allstar side
-	pass := len(logs) > 0
+	pass := len(logs) == 0
```
laurentsimon, line 185: do you need to filter by log 'type' (debug, warn)?

laurentsimon (approved):

pkg/policies/binary/binary.go
jeffmendoza: All the logs I have seen so far from binary artifacts has been actual artifacts. I'm calling checks.BinaryArtifacts(cr) directly. Will there be other logs inserted?

PR #194, opened 2022-06-01
Allow org-level configs to specify a base repo to merge configs

jeffmendoza:
Any org-level file can set "baseConfig" to a GitHub "owner/repo". Then the same
file in that repo will be read first, and the contents of the current file will
be applied on top of the baseConfig as a RFC7396 JSON merge patch:
https://datatracker.ietf.org/doc/html/rfc7396.


laurentsimon (approved):
Very cool!

pburkholder:
Does the satellite organization get to override any configuration setting, or just the ones in `allstar.yaml`?

Issue #170, opened 2022-04-14
Handle running out of quota / rate limit


jeffmendoza:
Allstar aggressively caches and saves the cache between runs, however it is still running out of quota on orgs with a large number of small repositories. For some reason the cache is not helping avoid this from run to run. We'll need to figure out what is going on with the cache, and probably implement a proper rate limit delay to pick up where we left off.

yorinasub17:
>  For some reason the cache is not helping avoid this from run to run.

I looked into this and traced it down to because of the ordering of the transports. The `httpcache` transport uses the request headers to determine whether or not to return the cached responses, and because the `httpcache` transport is run first, this is ALWAYS different.

Specifically, what happens is that the `httpcache` transport records the original request headers as new headers prefixed with `X-Varied-HEADER_KEY` headers on the response. However, this happens after the GH `RoundTrip` function is called, which means that when `httpcache` records the request headers, it includes the `Authorization` token from the underlying GH transport. Well this doesn't match the next time the cache lookup happens because the request checking happens BEFORE calling `RoundTrip` on the underlying transport, which means the request header is missing the `Authorization` token.

Once I reordered the transports, I was able to observe the cache being utilized much more. Here is the PR with this! https://github.com/ossf/allstar/pull/189

jeffmendoza:
Awesome, thanks for this! I'll try it out and see

jeffmendoza:
The above along with #197 seems to have helped here a lot, closing for now.

PR #195, opened 2022-06-01
List the repository for repo config folder before attempting lookup so that cache is utilized

yorinasub17:
This is a follow up to https://github.com/ossf/allstar/pull/192 to improve https://github.com/ossf/allstar/issues/170

When I introduced #192, I inadvertently significantly increased the number of API calls that need to be made, since now for each policy and repo combo, `allstar` needs to make 3 calls: the root org, org repo folder, and the repo `.allstar` folder. This increased the API calls by `5N` (5 policies) where `N` is the number of repos, making it constantly hit the rate limits.

To improve this, this PR adds a check to the config fetching routine that lists the contents of the org level config repo to determine if a folder for the target repo exists there before attempting to fetch it. Normally this would make the situation worse by adding an extra API call. In this case it helps a lot because the request to list the root is cache friendly. That is, that request will be cached and `allstar` will continuously use the cached response during the run across different repos and policies.

In effect, this helped us reduce the number of API calls Allstar made to Github from ~4400 to ~1800.

### DIFF
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -137,22 +137,49 @@ func fetchConfig(ctx context.Context, r repositories, owner, repoIn, name string
 	var p string
 	switch cl {
 	case OrgLevel, OrgRepoLevel:
+		var root string
 		_, rsp, err := r.Get(ctx, owner, operator.OrgConfigRepo)
 		if err == nil {
 			repo = operator.OrgConfigRepo
+			root = ""
 			p = name
 			if cl == OrgRepoLevel {
 				p = path.Join(repoIn, p)
 			}
 		} else if rsp != nil && rsp.StatusCode == http.StatusNotFound {
 			repo = githubConfRepo
+			root = operator.OrgConfigDir
+			p = path.Join(operator.OrgConfigDir, name)
 			if cl == OrgRepoLevel {
 				p = path.Join(operator.OrgConfigDir, repoIn, name)
-			} else {
-				p = path.Join(operator.OrgConfigDir, name)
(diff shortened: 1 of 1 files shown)


jeffmendoza:
I'm curious how getting the directory is more cache friendly than the file. I understand we won't have the file in cache if it is 404, but does GitHub conditional requests not work also?

jeffmendoza:
Did a little testing and confirmed that there is no etag in the 404 response from get repo or get contents. We probably want to abstract getcontents everywhere and have it walk the path from the root of the repo. I'm thinking we probably want to save whether an org has a '.allstar' repo across the enforce run as well.

yorinasub17:
> I'm curious how getting the directory is more cache friendly than the file. I understand we won't have the file in cache if it is 404, but does GitHub conditional requests not work also?

Ah so listing the directory is cache friendly within the loop as `allstar` is working through the repos, vs reusing the cache across multiple iterations of the loop.

The cache key is the URL being requested, so when we list the root directory of `.allstar`, that request is the same no matter what policy or repo we are evaluating. Which means that for all repos, this directory listing request can reuse the response from the first call, and thus avoid a call to the individual file if the repo directory doesn't exist in the first place.

OTOH, even if GitHub caches `404` requests, we would need to make that call for every repo the first time the loop is run to reuse the cache on the next loop, because the URL for the repo config is unique for each repo-policy pair.

Hope that makes sense!

yorinasub17:
Closing as superseded by https://github.com/ossf/allstar/pull/197

PR #197, opened 2022-06-01
Update logic in getting config files to avoid hitting 404s.

jeffmendoza:
Config files can reside in multiple places, and Allstar needs to look for them
for each repository. We have a local cache and use GitHub conditonal requests
to preserve quota. However, 404s don't work with cache or conditonal
requests. Therefore we should avoid 404s where possible. This change includes
two parts:

1. Preserve whether or not the ".allstar" or ".github" repositories exist
across an installation run. We can't avoid 404s if we don't do this. Each
config file lookup for each repo would need to do a get on both of these if
they don't exist.

2. When looking for a file, start at the repo top level, and get the directory
listing, then see if the file exists before trying to get the file. The
directory listing should exist at least at the top level, so we can cache that
and also use conditional requests.

### DIFF
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -127,38 +127,52 @@ const (
 	RepoLevel
 )
 
+var walkGC func(context.Context, repositories, string, string, string,
+	*github.RepositoryContentGetOptions) (*github.RepositoryContent,
+	[]*github.RepositoryContent, *github.Response, error)
+
+func init() {
+	walkGC = walkGetContents
+}
+
+type repositories interface {
+	Get(context.Context, string, string) (*github.Repository,
+		*github.Response, error)
+	GetContents(context.Context, string, string, string,
+		*github.RepositoryContentGetOptions) (*github.RepositoryContent,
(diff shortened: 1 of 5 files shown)


jeffmendoza:
@yorinasub17 I took #195 and expanded it, PTAL.
I'd like to lookup the dir listing in all cases, ex:

-  Only a  subset of policies setup in main .allstar repo (some files missing)
- No .allstar repo, and a .github repo with no allstar directory.
- No .allstar directory in each repository.

In those cases we should try to use the cached dir listing, and avoid the getcontents 404.

Also, if an org doesn't have a .allstar repo, we will 404 a get to that repo for each config file in each repo, we should avoid that as well.

laurentsimon (approved):

yorinasub17 (approved):
Makes sense to me! This should work for us.

jeffmendoza:
I'll monitor in staging and see how it is working =)

yorinasub17:
Just wanted to mention that I just deployed this new version into our environment and confirmed the API usage is within the limits for our needs, at 1500 requests for the run.

PR #203, opened 2022-06-08
New feature: Use private key from environment variable

markdboyd:
We are running our own instance of Allstar as operators for security reasons.

And in our case, we have a credential manager service that integrates directly with our CI/CD to deploy apps. This credential service doesn't provide any easy way to access credentials using a service provided by `runtimevar`.

Thus, it's much easier for me to just configure our CI to set the app private key directly as an environment variable when deploying Allstar. So I made some small additions to Allstar to support this behavior.

Thanks for making this great app! It's been so useful to our team already.

### DIFF
--- a/cmd/allstar/main.go
+++ b/cmd/allstar/main.go
@@ -33,6 +33,8 @@ import (
 
 func main() {
 	setupLog()
+	log.Info().
+		Msg("Starting application")
 	ctx, cf := context.WithCancel(context.Background())
 
 	ghc, err := ghclients.NewGHClients(ctx, http.DefaultTransport)
--- a/cmd/allstar/main.go
+++ b/cmd/allstar/main.go
@@ -33,8 +33,6 @@ import (
 
 func main() {
 	setupLog()
-	log.Info().
-		Msg("Starting application")
 	ctx, cf := context.WithCancel(context.Background())
 
 	ghc, err := ghclients.NewGHClients(ctx, http.DefaultTransport)
--- a/operator.md
+++ b/operator.md
@@ -37,7 +37,9 @@ may need to edit `pkg/ghclients/ghclients.go` and add a new import line for your
 secret service, ex: `_ "gocloud.dev/runtimevar/gcpsecretmanager"`.
 
 Alternatively, you can provide the AppID and KeySecret as environment variables
-`APP_ID` and `KEY_SECRET`.
+`APP_ID` and `KEY_SECRET`. Or, you can provide the private key directly in the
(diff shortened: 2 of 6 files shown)


jeffmendoza:
Thanks for the PR! Is your CD system able to place the secret in a file upon deployment? This could work as a "filevar" see the last example here: https://gocloud.dev/howto/runtimevar/#local

I know setting environment variables directly to secrets (vs files or volume mounts) is not especially recommended in container environments, but I'm not sure if that translates to all environments. ( https://youtu.be/j3QJRdiTr1I?t=582 )

That said I understand that this will likely be required by someone's environment, and am happy to include it. Let me think about how to position it in the docs and get back to you. We may want to include some remarks to recommend the file / secret manager method.

pburkholder:
Hi - I'm the security and compliance lead for cloud.gov. We need this in _our_ environment, it's not theoretical. I will not approve a system using a private key as file when it can be provided as an env var instead. I hope that makes sense.

pburkholder:
(I was hasty with that comment. Let me discuss with my team more, but: we'll still probably need this)

markdboyd:
@jeffmendoza Thanks for that video! It definitely makes good points about the pitfalls of secrets as environment variables, particularly the point about leaky logs. We're investigating the feasibility of using a local file to store the secret and will get back to you.

markdboyd:
@jeffmendoza It turns out that we can use `filevar` to load our secret from a local file. It requires some extra plumbing in our CD pipeline setup, but it works.

I agree with you that there are some definite tradeoffs to using environment variables for secrets vs some kind of credential store. Using Allstar would be easiest for us with the PR merged, but we can work around it if you're not comfortable including this change.

My only other thought is that in order to use `filevar`, I still have to modify `ghclients.go` and ultimately maintain that change in our fork. Ultimately, I'd love to get away from having a fork at all and be able to use your upstream repo directly. But I also understand that `filevar` is itself a utility more intended for development and testing, so perhaps not something you want to include as available by default in the code.

jeffmendoza:
> Using Allstar would be easiest for us with the PR merged, but we can work around it if you're not comfortable including this change.

I'm ok with merging. As a security project, I'd like it to encourage more secure practices, and discourage less secure ones. Some language around the trade-offs in the operator.md should accomplish that. I can work on that early next week.

> Ultimately, I'd love to get away from having a fork at all and be able to use your upstream repo directly. 

Yes, absolutely. I'm happy to add any extra imports for any `runtimevar` provider.

jeffmendoza:
Superseded by #207

PR #212, opened 2022-06-15
New feature: Add ability to check & fix whether signed commits for protected branches

markdboyd:
My organization is looking to audit is whether we require signed commits for merging to protected branches on our repos. This feature seemed to be missing from Allstar, so I've added it.

I also added the ability to auto-fix the signed commits setting for protected branches if it is out of sync with the expected setting. I can confirm that running an instance of Allstar with these changes and configured with `requireSignedCommits: true` in `branch_protection.yaml`, a repo was properly updated to enable "Require signed commits" on the branch protection settings.

I'm a little unsure about whether some of my error handling logic is correct, but so far everything is looking good in unit tests and manual testing.

### DIFF
--- a/pkg/policies/branch/branch.go
+++ b/pkg/policies/branch/branch.go
@@ -75,6 +75,9 @@ type OrgConfig struct {
 	// EnforceOnAdmins : set to true to apply the branch protection rules on
 	// administrators as well.
 	EnforceOnAdmins bool `json:"enforceOnAdmins"`
+
+	// RequireSignedCommits : set to true to require signed commits on protected branches, default false
+	RequireSignedCommits bool `json:"requireSignedCommits"`
 }
 
 // RepoConfig is the repo-level config for Branch Protection
@@ -116,6 +119,10 @@ type RepoConfig struct {
 	// specifying an empty list (`requireStatusChecks: []`) will override the
 	// setting to be empty.
 	RequireStatusChecks []StatusCheck `json:"requireStatusChecks"`
+
+	// RequireSignedCommits overrides the same setting in org-level, only if
+	// present.
+	RequireSignedCommits *bool `json:"requireSignedCommits"`
 }
 
 // StatusCheck is the config description for specifying a single required
@@ -146,6 +153,7 @@ type mergedConfig struct {
 	EnforceOnAdmins       bool
 	RequireUpToDateBranch bool
 	RequireStatusChecks   []StatusCheck
+	RequireSignedCommits  bool
 }
 
 type details struct {
@@ -156,6 +164,7 @@ type details struct {
 	EnforceOnAdmins       bool
 	RequireUpToDateBranch bool
 	RequireStatusChecks   []StatusCheck
+	RequireSignedCommits  bool
 }
 
 var configFetchConfig func(context.Context, *github.Client, string, string, string, config.ConfigLevel, interface{}) error
@@ -191,6 +200,10 @@ type repositories interface {
 		*github.Protection, *github.Response, error)
 	UpdateBranchProtection(context.Context, string, string, string,
 		*github.ProtectionRequest) (*github.Protection, *github.Response, error)
+	GetSignaturesProtectedBranch(ctx context.Context, owner, repo, branch string) (
+		*github.SignaturesProtectedBranch, *github.Response, error)
+	RequireSignaturesOnProtectedBranch(ctx context.Context, owner, repo, branch string) (
+		*github.SignaturesProtectedBranch, *github.Response, error)
 }
 
 // Check performs the polcy check for Branch Protection based on the
@@ -373,6 +386,17 @@ func check(ctx context.Context, rep repositories, c *github.Client, owner,
 				pass = false
 			}
 		}
+
+		signatureProtectionEnabled, err := getSignatureProtectionEnabled(ctx, rep, owner, repo, b)
+		if err != nil {
+			return nil, err
+		}
+		d.RequireSignedCommits = signatureProtectionEnabled
+		if mc.RequireSignedCommits && !d.RequireSignedCommits {
+			pass = false
+			text = text + fmt.Sprintf("Signed commits required, but not enabled for branch: %v\n", b)
+		}
+
 		ds[b] = d
 	}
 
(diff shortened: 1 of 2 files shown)


pkg/policies/branch/branch.go
```diff
@@ -568,6 +597,51 @@ func fix(ctx context.Context, rep repositories, c *github.Client,
...
+
+		updateRequiredSignatures := mc.RequireSignedCommits != *sp.Enabled
+		if updateRequiredSignatures {
+			var signatureStatus string
+			if mc.RequireSignedCommits && !*sp.Enabled {
+				signatureStatus = "required"
+				_, rsp, err = rep.RequireSignaturesOnProtectedBranch(ctx, owner, repo, b)
+			} else {
+				signatureStatus = "optional"
+				rsp, err = rep.OptionalSignaturesOnProtectedBranch(ctx, owner, repo, b)
```
markdboyd: It seemed counter-intuitive to me to write code that **disables** the requirement of signed commits on protected branches, but if `requireSignedCommits` is configurable, then the code should make sure that branch protection settings are in alignment?

markdboyd: I changed my mind. To me, the meaning of `requireSignedCommits` means that we expect the "Require signed commits" setting to be enabled for any protected branch. 

If `requireSignedCommits: true`, then we should enable that setting for any protected branch which doesn't have it enabled. 

If `requireSignedCommits: false`, then I think we just don't care about whether the protected branches have this setting enabled or not. We don't need to actively disable it if it is already enabled.

pkg/policies/branch/branch.go
```diff
@@ -356,6 +369,20 @@ func check(ctx context.Context, rep repositories, c *github.Client, owner,
 				pass = false
 			}
 		}
+
+		sp, _, err := rep.GetSignaturesProtectedBranch(ctx, owner, repo, b)
+		if sp != nil {
```
jeffmendoza: use the sp.GetEnabled accessor https://pkg.go.dev/github.com/google/go-github/v45/github#SignaturesProtectedBranch.GetEnabled

pkg/policies/branch/branch.go
```diff
@@ -356,6 +369,20 @@ func check(ctx context.Context, rep repositories, c *github.Client, owner,
 				pass = false
 			}
 		}
+
+		sp, _, err := rep.GetSignaturesProtectedBranch(ctx, owner, repo, b)
```
jeffmendoza: We want to do:
```
		if err != nil {
			return nil, err
		}
```
Here if we get an error, instead of the "could not determine" below. 

Also, we want to check for expected errors from GitHub, For example they may give a 404 if there is no protection on the branch so a `if rsp != nil && rsp.StatusCode == http.StatusNotFound` check should be done first and set `d.RequireSignedCommits` to `false` and continue on.

pkg/policies/branch/branch.go
```diff
@@ -568,6 +595,42 @@ func fix(ctx context.Context, rep repositories, c *github.Client,
...
+					Str("org", owner).
+					Str("repo", repo).
+					Str("area", polName).
+					Str("branch", b).
+					Msg("Fix action selected, but cannot access current commit signing settings.")
+				return nil
+			}
+			return err
+		}
+		if mc.RequireSignedCommits && !*sp.Enabled {
```
jeffmendoza: Same, use accessor

pkg/policies/branch/branch.go
```diff
@@ -568,6 +595,42 @@ func fix(ctx context.Context, rep repositories, c *github.Client,
 				Str("area", polName).
 				Msg("Updated with Fix action.")
 		}
+
+		sp, rsp, err := rep.GetSignaturesProtectedBranch(ctx, owner, repo, b)
+		if err != nil {
+			if rsp != nil && rsp.StatusCode == http.StatusForbidden {
+				log.Warn().
```
jeffmendoza: Usually we either warn and continue, or return a non-nil error, not both.

Both of these seem like errors that we should pass up, there should be at least a minimal branch protection setup on this branch at this point. If we want to add info to the error, we can wrap it.

jeffmendoza (changes requested):

pkg/policies/branch/branch.go
markdboyd: > Both of these seem like errors that we should pass up

Do you mean "pass up" in the call stack by returning them or "pass up" as in skip them by warning and continuing?

markdboyd: I was trying to mimic the pattern of error handling seen here, but that may have been naive

https://github.com/ossf/allstar/blob/6d2a249050ba0f2ef8453e7c73037f360728f652/pkg/policies/branch/branch.go#L583-L594

markdboyd: I assumed that you meant we should optionally wrap the errors and return them. But happy to fix if I have misunderstood.

pkg/policies/branch/branch.go
```diff
@@ -373,6 +386,23 @@ func check(ctx context.Context, rep repositories, c *github.Client, owner,
...
+				continue
+			}
+			return nil, err
+		}
+		if sp != nil {
+			d.RequireSignedCommits = sp.GetEnabled()
+		}
+		if mc.RequireSignedCommits && !d.RequireSignedCommits {
+			pass = false
+			text = text + fmt.Sprintf("Signed commits not required for branch %v\n", b)
```
jeffmendoza: How about "Signed commits required, but not enabled for branch: ..." ?

pkg/policies/branch/branch.go
```diff
@@ -373,6 +386,23 @@ func check(ctx context.Context, rep repositories, c *github.Client, owner,
 				pass = false
 			}
 		}
+
+		sp, rsp, err := rep.GetSignaturesProtectedBranch(ctx, owner, repo, b)
+		if err != nil {
+			if rsp != nil && rsp.StatusCode == http.StatusNotFound {
+				d.RequireSignedCommits = false
+				continue
```
jeffmendoza: Oops, I didn't mean literal `continue`, sorry! If we do that here, it will miss the check for pass below I think.

pkg/policies/branch/branch.go
```diff
@@ -598,6 +628,36 @@ func fix(ctx context.Context, rep repositories, c *github.Client,
 				Str("area", polName).
 				Msg("Updated with Fix action.")
 		}
+
+		sp, rsp, err := rep.GetSignaturesProtectedBranch(ctx, owner, repo, b)
+		if err != nil {
```
jeffmendoza, line 627: We want to check for 404 here as well (Like in the check func) and assume protection=false if so, then set it below if so.

jeffmendoza (changes requested):

pkg/policies/branch/branch.go
jeffmendoza: Yep, that looks good for the `RequireSignaturesOnProtectedBranch` as you have now.

pkg/policies/branch/branch.go
markdboyd: To centralize this logic and keep it consistent, I added a helper function `getSignatureProtectionEnabled` and some tests for it. I hope that wasn't overkill.

markdboyd:
@jeffmendoza I think this is ready for re-review. Thanks!

jeffmendoza (approved):

jeffmendoza:
@markdboyd Sorry for the delay, I was traveling last week. Looks great, thanks for the work! I will test this out in staging a bit then deploy to prod.

PR #213, opened 2022-06-15
Bump scorecard, and move to new check result to new interface.

jeffmendoza:


laurentsimon (approved):

pkg/scorecard/scorecard_test.go
```diff
@@ -51,6 +51,10 @@ func (m mockRC) ListBranches() ([]*clients.BranchRef, error) {
 	return nil, nil
 }
 
+func (m mockRC) GetBranch(branch string) (*clients.BranchRef, error) {
```
azeemshaikh38, line 54: Unrelated to the PR, Scorecard package has a MockRepoClient you can use instead of maintaining one yourself - https://github.com/ossf/scorecard/blob/main/clients/mockclients/repo_client.go

PR #214, opened 2022-06-15
Add configuration option to set enforce_admins for required status checks

yorinasub17:
This PR introduces a new configuration option for the branch protection policy to allow enforcing the "Include Administrators" flag on the branch protection rule (see attached screenshot):

Screen Shot 2022-06-15 at 5 06 18 PM

### DIFF
--- a/pkg/policies/branch/branch.go
+++ b/pkg/policies/branch/branch.go
@@ -71,6 +71,10 @@ type OrgConfig struct {
 	// order to merge into the protected branch. Each entry must specify
 	// the context, and optionally an appID.
 	RequireStatusChecks []StatusCheck `json:"requireStatusChecks"`
+
+	// EnforceOnAdmins : set to true to apply the branch protection rules on
+	// administrators as well.
+	EnforceOnAdmins bool `json:"enforceOnAdmins"`
 }
 
 // RepoConfig is the repo-level config for Branch Protection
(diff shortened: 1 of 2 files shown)


pkg/policies/branch/branch.go
```diff
@@ -455,6 +472,10 @@ func fix(ctx context.Context, rep repositories, c *github.Client,
 			EnforceAdmins:        p.EnforceAdmins.Enabled,
```
jeffmendoza, line 475: Need to set this above as well
```
				// No existing protection, create from config.
				afp := !mc.BlockForce
				pr := &github.ProtectionRequest{
					AllowForcePushes: &afp,
				}
```

pkg/policies/branch/branch.go
```diff
@@ -321,6 +330,14 @@ func check(ctx context.Context, rep repositories, c *github.Client, owner,
 				d.BlockForce = false
 			}
 		}
+		ea := p.GetEnforceAdmins()
+		d.EnforceOnAdmins = false
```
jeffmendoza:
```
d.EnforceOnAdmins = (ea != nil  && ea.Enabled)
```

jeffmendoza (changes requested):

pkg/policies/branch/branch.go
yorinasub17: Ah good catch! Fixed in 2cbc17cb002f77b3260b407bc8ca273fc14abb0d, and also added a regression test 15c4e0712cbe6094c1196d30cff836e25aefc825

pkg/policies/branch/branch.go
yorinasub17: Fixed in 2cbc17cb002f77b3260b407bc8ca273fc14abb0d

jeffmendoza (approved):

jeffmendoza:
Thanks!

Issue #159, opened 2022-04-05
When should issues be closed?


jeffmendoza:
Currently the GitHub issue code is only triggered when:

- Allstar is enabled
- The policy is enabled
- The policy action is `issue`.

At that point. Allstar will ensure an issue is open if the policy fails, and ensure the issue is closed if the policy passes.

In any other case, Allstar does not check the issue status (open or closed). This means, Allstar will leave issues open if:

- The policy or Allstar is disabled
- The action is changed to `log` or `fix`.
- The `issueRepo` is set, unset, or changed.
- Allstar is uninstalled. (Can never close in this case).

Should Allstar try to close issues in some or all of these cases? If so, which ones? Allstar doesn't keep state, so this means it will search for issues in all those cases whenever a repo is checked.

azeemshaikh38:
My 2 cents - let's say a policy is enabled with `issue` action and is seeing violations. The 3 possible updates the user could make:

1. Update the policy i.e disable or update some policy params (like add a binary artifact exception). In this case, we should ensure that the issue gets closed if the new policy params cause no violations (including the case where policy gets disabled). 
2. Update the action type. Let's say the action type changed from `issue` to `email`, it would makes sense to me if the old issue did not get auto-closed. `issue` -> `fix` is a special case of this so I'm ok if the issue is left open in this case too. Another reason for me saying this is that, since AllStar doesn't keep state, we will need to add it only to support this which may not be worth the effort.
3. Update the `issueRepo`. Same as (2), we can leave the older issue open since we won't be spamming it anyways.

When we make architectural changes to AllStar backend for scaling, we can consider keeping state so that we have better support for (2) and (3). 

@brianrussell2 fyi.

agonzalez-plume:
Interestingly, I just updated Allstar branchprotection mode from issue -> log while diagnosing excessive Github emails going to some people (not all emails were from Allstar). I had wondered if that would close the Issue or not. I guess it doesn't, which is what I expected. 

I do wonder if the priority for the Allstar check for a class of repos gets excluded or cancelled after initial configuration. If I optOut a class of repos, should that close the Issue? In that case, I believe it should close the issue.

markdboyd:
@jeffmendoza I was just wondering about this myself. My opinion is that we should update Allstar to close issues if the action is set to "fix". Currently, you could have this experience:

- Action is set to "issue". Allstar runs scans, finds repos out of compliance, and creates Github issues.
[...]

PR #218, opened 2022-06-16
Update Allstar to close issues when action: "fix"

markdboyd:
Related to #159 

Currently, this experience is possible:

- Action is set to "issue". Allstar runs scans, finds repos out of compliance, and creates Github issues
- Action is set to "fix" to automatically fix the failing repos, but the Github issues remain open unless you set action back to "issue" and wait for the scans to re-run

If action is set to "fix", then for a given repo I would like to see this behavior:

- On initial scan, issues with repo are automatically fixed
- Subsequent scan of the repo closes Github issues, if any

I can't really think of a downside this approach, so I went ahead and implemented it

### DIFF
--- a/pkg/enforce/enforce.go
+++ b/pkg/enforce/enforce.go
@@ -248,7 +248,7 @@ func RunPolicies(ctx context.Context, c *github.Client, owner, repo string, enab
 					Msg("Unknown action configured.")
 			}
 		}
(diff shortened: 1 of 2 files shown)


jeffmendoza (approved):

Jung47:
authored

PR #219, opened 2022-06-17
Bug: Fix issue with error scope when running EnforceAll

markdboyd:
While running scans to fix my branch protection failures with `action: fix`, I noticed that the scan just stopped abruptly at a certain repo, but there was no log (even with `ALLSTAR_LOG_LEVEL: "debug"`) to indicate what had happened.

After some investigation, I noticed that the error logging in `EnforceAll` seemed like it was swallowing the error. Basically, inside the loop that iterates over app installation repos, [the code is setting the error when `RunPolicies` fails on a repo and if the error is not nil it breaks out of the loop](https://github.com/ossf/allstar/blob/ec33a2bdb671731dffa55ab3f557f6d635b6f070/pkg/enforce/enforce.go#L109-L112). [But the code does not try to log the error until outside of the loop](https://github.com/ossf/allstar/blob/ec33a2bdb671731dffa55ab3f557f6d635b6f070/pkg/enforce/enforce.go#L125-L130). I think this does not work because the scope of the value assigned to `err` on line 109 is only within the `for` loop.

I can confirm that after making the changes in this PR and redeploying, I can see the error logged that caused my scan to fail. Incidentally, it's related to the changes in #212, so there could be more changes coming in that PR. 

I did some light refactoring of the code to make it easier to test that an error is definitely returned when encountered looping over the repos. And I did some test cleanup as well.

### DIFF
--- a/pkg/enforce/enforce.go
+++ b/pkg/enforce/enforce.go
@@ -38,8 +38,9 @@ var issueClose func(ctx context.Context, c *github.Client, owner, repo, policy s
 var getAppInstallations func(ctx context.Context, ghc ghclients.GhClientsInterface) ([]*github.Installation, error)
 var getAppInstallationRepos func(ctx context.Context, ghc ghclients.GhClientsInterface, ic *github.Client) ([]*github.Repository, *github.Response, error)
 var isBotEnabled func(ctx context.Context, c *github.Client, owner, repo string) bool
+var runPolicies func(ctx context.Context, c *github.Client, owner, repo string, enabled bool) (EnforceRepoResults, error)
 
-type EnforceResults = map[string]bool
+type EnforceRepoResults = map[string]bool
 type EnforceAllResults = map[string]map[string]int
 
 func init() {
@@ -49,6 +50,7 @@ func init() {
 	getAppInstallations = getAppInstallationsReal
 	getAppInstallationRepos = getAppInstallationReposReal
 	isBotEnabled = config.IsBotEnabled
+	runPolicies = RunPolicies
 }
 
 // EnforceAll iterates through all available installations and repos Allstar
@@ -60,7 +62,6 @@ func init() {
 func EnforceAll(ctx context.Context, ghc ghclients.GhClientsInterface) (EnforceAllResults, error) {
 	var repoCount int
 	var enforceAllResults = make(EnforceAllResults)
-
 	insts, err := getAppInstallations(ctx, ghc)
 	if err != nil {
 		return nil, err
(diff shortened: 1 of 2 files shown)


markdboyd:
The actual error I was seeing is documented/fixed in #220

pkg/enforce/enforce.go
```diff
@@ -40,8 +40,9 @@ var getAppInstallations func(ctx context.Context, ghc ghclients.GhClientsInterfa
 var getAppInstallationRepos func(ctx context.Context, ghc ghclients.GhClientsInterface, ic *github.Client) ([]*github.Repository, *github.Response, error)
 var isBotEnabled func(ctx context.Context, c *github.Client, owner, repo string) bool
 
-type EnforceResults = map[string]bool
+type EnforceRepoResults = map[string]bool
 type EnforceAllResults = map[string]map[string]int
+type RunPoliciesFunc func(ctx context.Context, c *github.Client, owner, repo string, enabled bool) (EnforceRepoResults, error)
```
jeffmendoza: Lets use a global var here like the other test-injected functions.
```
var runPolicies ...
init() { runPolicies = runPoliciesReal }
```

pkg/enforce/enforce.go
```diff
@@ -139,6 +127,34 @@ func EnforceAll(ctx context.Context, ghc ghclients.GhClientsInterface) (EnforceA
 	return enforceAllResults, nil
 }
 
+func RunPoliciesOnInstRepos(ctx context.Context, repos []*github.Repository, ghclient *github.Client, runPolicies RunPoliciesFunc) (
```
jeffmendoza: Can this be non-exported?

pkg/enforce/enforce.go
```diff
@@ -139,6 +127,34 @@ func EnforceAll(ctx context.Context, ghc ghclients.GhClientsInterface) (EnforceA
...
+		for policyName, passed := range enforceResults {
+			if !passed {
+				if instResults[policyName] == nil {
+					instResults[policyName] = make(map[string]int)
+				}
+				instResults[policyName]["totalFailed"] += 1
+			}
+		}
+	}
+	return instResults, owner, repoLoopErr
```
jeffmendoza: Add `config.ClearInstLoc(owner)` here, then remove owner from the return values.

jeffmendoza (changes requested):

jeffmendoza (approved):

jeffmendoza:
Awesome, thanks!

PR #229, opened 2022-06-29
Implement org-level Outside Collaborators exemptions (recreated)

### LINKED ISSUE #223: Suggestion: Support allowlisting an individual to be an outside contributor
Currently the only mechanism to allow an outside individual to have admin/push access to a repo is by turning off those policy configurations, or opting the repository out.  There should be a way to specify a map of repositories to GitHub users which are allowed to be administrators and/or have push access to a repo.

ghost:
Allows org-level configuration of exemptions to the Outside Collaborators policy.

Example `outside.yaml`:
```yaml
optConfig:
  optOutStrategy: true
action: issue
exemptions:
  - user: test-user2134
    repo: problematic
    push: true
    admin: false
  - user: big-deal-user
    repo: *
    push: true
    admin: true
```

Resolves #223.


ghost:
Add glob support to repo option.

jeffmendoza:
@JonZeolla Does the proposed config above satisfy how you would use this feature?

pkg/policies/outside/outside.go
```diff
@@ -102,6 +112,24 @@ type mergedConfig struct {
 	PushAllowed             bool
 	AdminAllowed            bool
 	TestingOwnerlessAllowed bool
+	Exemptions              OutsideExemptions
+}
+
+type OutsideExemptions []*OutsideExemption
```
jeffmendoza: I'd just leave this as `[]*OutsideExemption` in the org config.

pkg/policies/outside/outside.go
```diff
@@ -293,12 +321,28 @@ func getUsers(ctx context.Context, r repositories, owner, repo, perm,
...
-			rv = append(rv, u.GetLogin())
+			if !exemptions.isExempt(repo, u.GetLogin(), perm) {
+				rv = append(rv, u.GetLogin())
+			}
 		}
 	}
 	return rv, nil
 }
 
+func (ee OutsideExemptions) isExempt(repo, user, access string) bool {
```
jeffmendoza: In general allstar is coded as procedural and not object oriented. In this case there is no benefit to this being a function on the type, it could just be a parameter.

pkg/policies/outside/outside.go
```diff
@@ -293,12 +321,28 @@ func getUsers(ctx context.Context, r repositories, owner, repo, perm,
...
+	for _, e := range ee {
+		if !(((e.Push || e.Admin) && access == "push") || (e.Admin && access == "admin")) {
+			continue
+		}
+		if g, err := glob.Compile(e.Repo); err == nil {
```
jeffmendoza: However, reading the docs for this library, compiling is the more expensive operation and matching is cheaper, so It might make sense to have a custom type saved somewhere (merged config?) that includes saving the glob.Glob for each entry. In that case, keeping this as a function of that custom type makes sense.

jeffmendoza (changes requested):
To sum up. Let's have the org config have `Exemptions []*OutsideExemption`, then have another type that contains the exemption structs along with saving the glob.Glob. You could either compile all the globs in `mergeConfig` or lazy-initialize them in `isExempt`

pkg/policies/outside/outside.go
ghost: It would make sense, but I'm unsure about where exactly to persist the globs so that they can be used for each repo in an org because the configs used within the outside package seem to be recreated for each repo. Open to suggestions about this!

jeffmendoza: Compiling once per-repo is better than repo * users. We could go with that, or stick to how it is now (not pre-optimitze) and make the function a standard function without a custom type and take `[]*OutsideExemption` as a parameter.

ghost: Ok, I put the cache into the outside OrgConfig.

jeffmendoza (approved):

PR #230, opened 2022-07-01
Allow the scheduling of periods during which issues aren't created

### LINKED ISSUE #201: Consider not pinging issues on the weekend
For example https://github.com/google/makerspace-auth/issues/47 -- it's existed in this state for years, and I get github notifications on my personal devices.  This is not kind to work-life balance, if this is intended for work projects...

ghost:
Enables users to define which actions should be employed during a period of time each week (max 3 days per week) at either the organization or repo level.

Example `allstar.yaml` config using this feature:
```yaml
optConfig:
  optOutStrategy: true
schedule:
  timezone: America/Los_Angeles
  actions:
    issue: true
    ping: false
  days:
    - saturday
    - sunday
```

The above config disables Allstar's automated commenting on issues on Saturday and Sunday in the `America/Los_Angeles` timezone.

Resolves #201.


ghost:
Move schedule to allstar top-level option, not part of optConfig

ethanent:
Hmm, perhaps it would be nice to differentiate between the creation of issues and comments. Right now in this PR, there is just the issue option which disables both, but perhaps it should only disable comments, with a separate option for creating new issues

ghost:
I want to refactor schedule into its own package to make testing nicer and future actions easy to add.

I'll mark the PR ready once that is done.

ghost:
The refactor is complete and the PR is ready for review. The signature of the issue Ensure function has been changed.

ghost:
The refactor has a number of advantages.
- Org and repo configs will contain Schedules, complete with the ShouldPerform method, ready for use in any future action
- Tests for the issue package are no longer using roundabout solutions to try different days because Ensure now has a time.Time argument
- Schedule package has its own testing and is straightforward to test

pkg/config/schedule/schedule.go
```diff
@@ -0,0 +1,111 @@
...
+// Package schedule provides the ActionSchedule type for use in config and
+// when deciding whether to perform an action.
+package schedule
+
+import (
+	"strings"
+	"time"
+)
+
+type ScheduleAction string
```
jeffmendoza: Remove `Actions  ScheduleConfigActions` and this type and code. Assume/hardcode that the schedule is for "notify" actions (issue/ping), and that log/fix will always work.

Overall I'd like to follow the YAGNI principle and make things simple for the sake of code and reduce the complexity for the user setting up the config file.

pkg/config/schedule/schedule.go
```diff
@@ -0,0 +1,111 @@
...
+	ScheduleActionFix         ScheduleAction = "fix"
+)
+
+// actionOverrides specifies, for each action A, a set of actions B.
+// If any action in B is disabled, action A will be disabled.
+var actionOverrides = map[ScheduleAction][]ScheduleAction{
+	ScheduleActionIssuePing: {ScheduleActionIssueCreate},
+}
+
+type ScheduleConfig struct {
```
jeffmendoza: Move this struct to the config package. I'd like all the structs that define what goes into `allstar.yaml` in one place.

pkg/config/schedule/schedule.go
```diff
@@ -0,0 +1,111 @@
...
+	"tuesday":   time.Tuesday,
+	"wednesday": time.Wednesday,
+	"thursday":  time.Thursday,
+	"friday":    time.Friday,
+	"saturday":  time.Saturday,
+}
+
+// ShouldPerform
+// The error may be ignored for default create behavior.
+func (sch *ScheduleConfig) ShouldPerform(a ScheduleAction, at time.Time) (bool, error) {
```
jeffmendoza: This can take the `ScheduleConfig` as a parameter

pkg/issue/issue.go
```diff
@@ -81,19 +82,22 @@ func getPolicyIssue(ctx context.Context, issues issues, owner, repo, policy, tit
 // Ensure ensures an issue exists and is open for the provided repo and
 // policy. If opening, re-opening, or pinging an issue, the provided text will
 // be included.
-func Ensure(ctx context.Context, c *github.Client, owner, repo, policy, text string) error {
-	return ensure(ctx, c, c.Issues, owner, repo, policy, text)
+func Ensure(ctx context.Context, c *github.Client, owner, repo, policy, text string, at time.Time) error {
+	return ensure(ctx, c, c.Issues, owner, repo, policy, text, at)
```
jeffmendoza: Remove the `at` parameter, and all the changes to the enforce package. We shouldn't change the package contract for testability. Create a private gloabal function pointer variable `timeNow` and set it to `time.Now` in the init function. Then in the tests, override the timeNow variable to a closure that returns the desired time for the test.

jeffmendoza (changes requested):

pkg/config/config.go
```diff
@@ -110,6 +116,32 @@ type RepoOptConfig struct {
...
+	// Set an action to false to disable it during scheduled period.
+	Actions ScheduleConfigActions `json:"actions"`
+
+	// Days specifies weekdays during which to apply the action toggles.
+	// eg. "saturday" or "sunday"
+	Days []string `json:"days"`
+}
+
+// ScheduleConfigActions is a set of actions to enable or disable.
+type ScheduleConfigActions struct {
```
jeffmendoza: Remove this type completely.

pkg/issue/issue.go
```diff
@@ -92,8 +95,12 @@ func ensure(ctx context.Context, c *github.Client, issues issues, owner, repo, p
 	if err != nil {
 		return err
 	}
-	if issue == nil {
-		oc, _, _ := configGetAppConfigs(ctx, c, owner, repo)
+	now := timeNow()
```
jeffmendoza: Actually, if we are only using now here to pass into `schedule.ShouldPerform`, just remove the time as a parameter to `schedule.ShouldPerform` and have it call timeNow() directly (mocked for testing).

pkg/config/schedule/schedule.go
```diff
@@ -0,0 +1,97 @@
...
+		falsebool := false
+		sch.Actions.Ping = &falsebool
+	}
+	// If action queried is always allowed by schedule, return true
+	if action == ScheduleActionIssueCreate && (sch.Actions.Issue == nil || *sch.Actions.Issue) {
+		return true, nil
+	}
+	if action == ScheduleActionIssuePing && (sch.Actions.Ping == nil || *sch.Actions.Ping) {
+		return true, nil
+	}
```
jeffmendoza, line 49: 52-63 can be removed without action parameter and sch.Actions.

pkg/issue/issue.go
jeffmendoza: Package issue shouldn't worry about the time, it should just be unit tested to call `schedule.ShouldPerform` and only create/ping if that returns true.

pkg/issue/issue.go
```diff
@@ -92,8 +95,12 @@ func ensure(ctx context.Context, c *github.Client, issues issues, owner, repo, p
...
+	oc, orc, rc := configGetAppConfigs(ctx, c, owner, repo)
+	osc := schedule.MergeSchedules(oc.Schedule, orc.Schedule, rc.Schedule)
+	createIssue, _ := schedule.ShouldPerform(osc, schedule.ScheduleActionIssueCreate, now)
+	pingIssue, _ := schedule.ShouldPerform(osc, schedule.ScheduleActionIssuePing, now)
```
jeffmendoza: Replace 101-102 with
```
if !schedule.ShouldPerform(osc) {
return nil
}
```
Then remove `createIssue` and `pingIssue`

pkg/config/schedule/schedule.go
```diff
@@ -0,0 +1,97 @@
...
+	// Get the day in timezone specified or default "" => UTC
+	loc, err := time.LoadLocation(sch.Timezone)
+	if err != nil {
+		return true, err
```
jeffmendoza: Just eat the error here and log a warning that the timezone was malformed, then ShouldPerform can return a bool only.

jeffmendoza (changes requested):

pkg/config/schedule/schedule.go
```diff
@@ -0,0 +1,86 @@
...
+	if sch == nil {
+		return true
+	}
+	// Get the day in timezone specified or default "" => UTC
+	loc, err := time.LoadLocation(sch.Timezone)
+	if err != nil {
+		log.Warn().
+			Str("tzstring", sch.Timezone).
+			Msg("Failed to load malformed timezone.")
+		return false
```
jeffmendoza: Let's return true here, so that we still get notifications on a mis-config.

jeffmendoza (changes requested):
Also fix lint.

jeffmendoza (approved):

PR #239, opened 2022-07-14
Olivekl fix quickstart instructions

olivekl:
Rewrite Quickstart Install instructions (in response to feedback that users were clicking the links without finishing the sentence and missing the instructions directly after them).
Add list format so users (hopefully) see all the instructions.

### DIFF
--- a/README.md
+++ b/README.md
@@ -72,7 +72,7 @@ Allstar is highly configurable. There are three main levels of controls:
    -  most repositories, except some that are opted out; 
    -  just a few repositories that are opted in. 
 
-These configurations are done in the organization's .`allstar` repository.
+These configurations are done in the organization's `.allstar` repository.
 
 - **Repo level:** Repository maintainers in an organization that uses
    Allstar can choose to opt their repository in or out of organization-level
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -171,12 +171,20 @@ policy violations by filing an issue. This is the quickest and easiest way to st
...
+1.  Install the Allstar app
+    1.  [Open the installation
+        page](https://github.com/apps/allstar-app) and click Configure
+    1.  If you have multiple organizations, select the one you want to
+        install Allstar on
+    1.  Select "All Repositories" under Repository Access, even if you
+        plan to disable Allstar on some repositories later
+1.  Fork the sample repository
+    1.  [Open the sample repository](https://github.com/jeffmendoza/dot-allstar-quickstart)
+        and click "Fork" button in the top right corner
```
jeffmendoza: I changed the sample repo to a template a little while ago. I believe the button is now "Use this template". Then for the next step, the name field is empty and the user just types in `.allstar`. Then click "Create repository from template"

jeffmendoza (changes requested):

README.md
olivekl: Looks good, thanks. Updated.

Do users need to keep the repo set as public? That's an option below the repo name field, and we should specify if so.

jeffmendoza (dismissed):

README.md
jeffmendoza: They don't, it is up to them

jeffmendoza (approved):

PR #240, opened 2022-07-19
Generic Scorecard Policy

jeffmendoza:
New policy that will run any scorecard check.

### DIFF
--- a/README.md
+++ b/README.md
@@ -313,6 +313,30 @@ the [GitHub
 tab](https://docs.github.com/en/code-security/getting-started/adding-a-security-policy-to-your-repository)
 that helps you commit a security policy to your repository.
 
+### Dangerous Workflow
+
+This policy's config file is named `dangerous_workflow.yaml`, and the [config
+definitions are
+here](https://pkg.go.dev/github.com/ossf/allstar/pkg/policies/workflow#OrgConfig).
+
+This policy checks the GitHub Actions workflow configuration files
+(`.github/workflows`), for any patterns that match known dangerous
+behavior. See the [Security Scorecards
+Documentation](https://github.com/ossf/scorecard/blob/main/docs/checks.md#dangerous-workflow)
+for more information on this check.
+
+### Generic Scorecard Check
+
+This policy's config file is named `scorecard.yaml`, and the [config definitions
+are
+here](https://pkg.go.dev/github.com/ossf/allstar/pkg/policies/scorecard#OrgConfig).
+
+This policy runs any scorecard check listed in the `checks` configuration. All
+checks run must have a score equal or above the `threshold` setting. Please see
+the [Security Scorecards
+Documentation](https://github.com/ossf/scorecard/blob/main/docs/checks.md)
+for more information on each check.
+
 ### Future Policies
 
 - Ensure dependabot is enabled.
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -313,6 +313,30 @@ the [GitHub
...
+are
+here](https://pkg.go.dev/github.com/ossf/allstar/pkg/policies/scorecard#OrgConfig).
+
+This policy runs any scorecard check listed in the `checks` configuration. All
+checks run must have a score equal or above the `threshold` setting. Please see
+the [Security Scorecards
+Documentation](https://github.com/ossf/scorecard/blob/main/docs/checks.md)
+for more information on each check.
+
 ### Future Policies
```
ghost, line 340: More checks from Scorecard can be taken off list!

ghost (commented):
Add policy to policies?

pkg/policies/scorecard/scorecard.go
```diff
@@ -0,0 +1,308 @@
...
+		return nil, err
+	}
+
+	var notify string
+	pass := true
+	f := make(map[string][]string)
+
+	for _, n := range mc.Checks {
+
+		l := checker.NewLogger()
```
ghost, line 158: Is it possible to move everything from 148 to 173 into its own function set in init so that you can test more thoroughly? You could also make a getScorecardClient function so you can create a nil scc while testing as well.

Something vaguely like:
```go
// testing version can return nil, nil
var getScorecardClient func() (*scorecard.ScClient, error)

var checkAndResult func(scc *scorecard.ScClient) (int, []string, error)
```

pkg/policies/scorecard/scorecard.go
```diff
@@ -0,0 +1,308 @@
...
+	// "registerCheck". See the check code for each name:
+	// https://github.com/ossf/scorecard/tree/main/checks For example, the name
+	// for the Signed Releases check is "Signed-Releases".
+	Checks []string `json:"checks"`
+
+	// Threshold is the score threshold that checks must meet to pass the
+	// policy. If all checks score equal or above the threshold, the Allstar
+	// policy will pass. The default is checker.MaxResultScore:
+	// https://pkg.go.dev/github.com/ossf/scorecard/v4@v4.4.0/checker#pkg-constants
+	Threshold int `json:"threshold"`
```
ghost, line 57: Allow a threshold to be set for each check? Is there a use case for this?

pkg/policies/scorecard/scorecard.go
```diff
@@ -0,0 +1,308 @@
...
+				Err(res.Error).
+				Msg("Scorecard check errored.")
+			break
+		}
+
+		logs := convertLogs(l.Flush())
+		if len(logs) > 0 {
+			f[n] = logs
+		}
+		if res.Score < mc.Threshold {
```
ghost: Handle InconclusiveResultScore?

pkg/policies/scorecard/scorecard.go
jeffmendoza: I ended up leaving this, but did add a test for Check.

pkg/policies/scorecard/scorecard.go
jeffmendoza: Don't want to overcomplicate now. If we have a usecase then we can add it.

ghost (approved):

PR #256, opened 2022-08-15
Parallize enforcements of installations 5x.

jeffmendoza:
In order to address performance, run 5 installations at the same time. Each
installation is still synchronous, so this should not affect quotas or enable
simultaneous requests for one installation.


ghost (approved):
LGTM! *

PR #260, opened 2022-08-17
Update issue body if body has changed

### LINKED ISSUE #251: Update issue when NotifyText changes
Currently, Allstar will reply periodically to an issue with body text, but maybe it should also update the issue body if a policy Result's NotifyText isn't the same as the corresponding issue body. This will be helpful for policies that have multiple steps to resolution, eg. Actions.

ghost:
Resolves #251.

Update the issue body if any value that goes into creating the body has changed, including:
- text (NotifyText from a policy result within enforce package)
- footer
- the issue body format itself


jeffmendoza:
Can you update README.md#actions with a description of the new behavior? Why did you choose to edit, vs reply? What happens to the ping replies?

PR #262, opened 2022-08-18
Update runPoliciesReal to continue with other policies on policy Check error

ghost:


jeffmendoza:
A check erroring should mean that we want to abort enforcement on the whole installation. Checks should eat/log anything that could be expected.

PR #268, opened 2022-08-23
Update issue if NotifyText has changed (second iteration)

ghost:
There were a few options to enable Allstar to detect result NotifyText changes:
- Keep track of issue state in an Allstar database so Allstar can compare NotifyText to issue text there
- Use GitHub list issue comments endpoint in loop to find last Allstar comment for comparison
- Maintain state within issue body

I picked the third because it results in fewer API calls and also allows Allstar to provide a link to the latest result update comment right in the issue body. Let me know if you think the complexity tradeoff (as compared to the second option) is worth it.

Two comment tags are added to the issue body to define section in which to place the current text hash comment and potentially update comment link. Issues created previously (without these tags) should not be affected by this change.

Additionally, getPolicyIssue was updated to only load issues as necessary to find the one it's looking for.

### DIFF
--- a/pkg/issue/issue.go
+++ b/pkg/issue/issue.go
@@ -17,8 +17,11 @@ package issue
 
 import (
 	"context"
+	"crypto/sha256"
+	"encoding/hex"
 	"fmt"
 	"net/http"
+	"strings"
 	"time"
 
 	"github.com/ossf/allstar/pkg/config"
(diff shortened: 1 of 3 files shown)


pkg/issue/issue.go
```diff
@@ -32,6 +35,11 @@ import (
 const issueRepoTitle = "Security Policy violation for repository %q %v"
 const sameRepoTitle = "Security Policy violation %v"
 
+const issueSectionHeaderFormat = "<!-- Edit section #%s -->"
+const resultTextHashCommentFormat = "<!-- Current result text hash: %s -->"
+const updateWarningFormat = "\n%s\n:warning: There is an updated version of this policy result! [Click here to see the latest update](%s)\n\n---\n\n"
```
ghost, line 40: Nit: remove "Click here"?

jeffmendoza (approved):

PR #286, opened 2022-09-02
Actions: use commit graph for version comparisons

ghost:
Convert tags and branches to commit hash, then, to compare, BFS graph until match found if the comparison operator isn't equals.

Notes about error handling:
- Require rules are skipped on error as before (eg. if version comparison ">= v3.0.1" used but there is no v3.0.1 tag)
- Deny rules are skipped on error as before (eg. if version comparison "<= v2.3.0" used but there is no v2.3.0 tag)
    - "v2.3.0" isn't considered equivalent to "2.3.0" beginning with this PR because tags are mapped to commits for comparison

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -3,7 +3,6 @@ module github.com/ossf/allstar
 go 1.16
 
 require (
-	github.com/Masterminds/semver/v3 v3.1.1
 	github.com/bradleyfalzon/ghinstallation/v2 v2.1.0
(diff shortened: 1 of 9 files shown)


ghost:
Add option `qualifyUntaggedCommits` for allow / require rule?

Motiff2 (approved):

justaugustus:
Going to close this as we haven't made a decision on this in a few years.
Thanks for raising this contribution though, @ethanent!

Issue #238, opened 2022-07-14
Proposal: Contributor Ladder


jeffmendoza:
This is a rough draft for discussion. We can also adapt this to Scorecards with minor changes, then share a generalized version at the foundation level.

| Name | Pre-reqs: Indicators to look for before promotion | Expectations | Recognition, Access, Effect of access |
| - | - | - | - |
| Contributor | One of: 1. Regular PRs for code/docs. 2. Regular feedback on discussion topics in issues/slack 3. Regular meeting attendance and contribution.  For a period of time… (30-60 days?) | Continue pre-req. | Added as GitHub Outside Collaborator to repo. (read or triage: project specific?) Tests on PRs run automatically without approval |
| Approver | Regular PRs for code/docs of moderate to substantial impact. Participates in many high-level proposals and discussions. | Available to have PRs assigned to them for review: Will review within 1 business day, otherwise notifies team when unable to review (ooo, etc.) Responds to requests for input/discussion over issues and slack | Added as GitHub Outside Collaborator to repo with write/push access. Approves and merges PRs |
| Maintainer | Consistently drives the project through code, proposals, project direction, leadership, etc. | Responds and makes final decisions on all large proposals, features, architecture changes, design, documentation of project | Full access and ownership. Creates releases, and finalizes changelog updates, sends project update announcements. |

Some notes:
- I’m only including "outside collaborator” access, as I have no idea the requirements to be added to the openssf org. **AI**: Figure out if we can incorporate org member at some level.
- Process for adding/removing members of levels is not covered, Will be handled by maintainer until a more formal process is needed.

jeffmendoza:
Draft for recommended OpenSSF level ladder, to propose

| Name | Pre-reqs: Indicators to look for before promotion | Expectations | Recognition, Access, Effect of access |
| - | - | - | - |
| Contributor | One of: 1. Regular PRs for code/docs. 2. Regular feedback on discussion topics in issues/slack 3. Regular meeting attendance and contribution.  For a period of time…  | Continue pre-req. | Added as GitHub Outside Collaborator to repo with read access. Tests on PRs run automatically without approval |
| Approver | Regular PRs for code/docs of moderate to substantial impact. Participates in many high-level proposals and discussions. | Available to have PRs assigned to them for review: Will review in a timely manner, otherwise notifies team when unable to review (ooo, etc.) Responds to requests for input/discussion over issues and slack | Added as GitHub Outside Collaborator to repo with write/push access. Approves and merges PRs |
[...]

PR #339, opened 2023-01-25
Update contribution guidelines.

jeffmendoza:
Includes a starter contributing.md, contributor ladder, code of conduct.


jeffmendoza:
@olivekl Also, please acknowledge that I'm putting you down as doc maintainer on the contributor ladder =) Thanks.

jeffmendoza:
Fixes #238

olivekl:
> @olivekl Also, please acknowledge that I'm putting you down as doc maintainer on the contributor ladder =) Thanks.

Ack, and apologies for the delay. I will review soon!

naveensrinivasan (dismissed):
Love it!

code-of-conduct.md
```diff
@@ -0,0 +1,128 @@
+# Contributor Covenant Code of Conduct
```
david-a-wheeler: Instead of this, grab or link to the OpenSSF CoC: https://openssf.org/community/code-of-conduct/

david-a-wheeler: Maybe add, at the end, a link to the OpenSSF COC?

jeffmendoza:
@olivekl Any thoughts on this? Thanks

SecurityCRob:
Why is a parallel Code of Conduct being developed?  I'm fine with a group self-managing and laying out how they want their collaborators/contributors to "level up", but as an OpenSSF project we should not set up a bifurcated CoC for a workgroup or project.  It all should align point to the Foundation's CoC.  If the OSSF has a gap, or could be improved by some additional ideas, let's amend that and not start up something else please.

jeffmendoza:
@SecurityCRob The CoC meant to be a copy/paste. I grabbed it from another OpenSSF repo. I'll double check that it is in fact the same one and add a link.

AevaOnline:
First of all, I'm very glad to see contribution guidelines getting updated.

I agree with previous comments about the CoC, and would like to add, please do not copy-paste it as this will lead to bit-drift if the original is ever changed. Please simply link to the top-level CoC.

I am concerned about a few aspects of this contributor ladder:
- there is not a clear description of the project's decision-making process, but the "expectations" column for the "leader" role appears to indicate that the leader "makes final decisions on all large proposals" and is, thus, essentially a [benevolent dictator](https://en.wikipedia.org/wiki/Benevolent_dictator_for_life). As a general rule, I (and I believe this to be true org-wide) prefer consensus-driven models with rotating leadership over time.
- the "leader" tier appears to be singular. Is that intentional? If not, I'd like to see this PR outline the path to elect new leaders, define term-length and electorate, etc.
- basically, the contributor guide should describe how someone can "climb" the ladder.

Lastly, I would like to point out that the OpenSSF definition of a sandbox project requires it to have at least two maintainers from different organizations, and the proposed set of maintainers does not appear to meet that bar. While we didn't expressly comment on "partial maintainership" (e.g., having someone only be a maintainer on a subset of a project, like docs) in PR112, it was not my intent to support that scenario in the OpenSSF's project progression policy. 

Is there someone else within the project who can share the highest tier today?

jeffmendoza:
Thanks for the feedback!

> I agree with previous comments about the CoC, and would like to add, please do not copy-paste it as this will lead to bit-drift if the original is ever changed. Please simply link to the top-level CoC.

Updated


> * there is not a clear description of the project's decision-making process, but the "expectations" column for the "leader" role appears to indicate that the leader "makes final decisions on all large proposals" and is, thus, essentially a [benevolent dictator](https://en.wikipedia.org/wiki/Benevolent_dictator_for_life). As a general rule, I (and I believe this to be true org-wide) prefer consensus-driven models with rotating leadership over time.
> * the "leader" tier appears to be singular. Is that intentional? If not, I'd like to see this PR outline the path to elect new leaders, define term-length and electorate, etc.

Pluralized the titles, added some group/consensus description to the leader(s) role.

> * basically, the contributor guide should describe how someone can "climb" the ladder.

Elaborated on the promotion process.

> Is there someone else within the project who can share the highest tier today?

Not currently, hoping that this ladder will help in this area.

olivekl:
> @olivekl Also, please acknowledge that I'm putting you down as doc maintainer on the contributor ladder =) Thanks.

Unfortunately, I don't think I can take on a docs maintainer role. I look forward to contributing when possible, but I don't think I can make a larger commitment at this time.

PR #341, opened 2023-02-03
Repo-level IgnorePaths isn't merged for the binary_artifacts check

coheigea:
I found it wasn't possible to add an exemption using IgnorePaths, because it's never merged in the config.

### DIFF
--- a/pkg/policies/binary/binary.go
+++ b/pkg/policies/binary/binary.go
@@ -279,7 +279,6 @@ func mergeConfig(oc *OrgConfig, orc, rc *RepoConfig, repo string) *mergedConfig 
 	mc := &mergedConfig{
 		Action:      oc.Action,
 		IgnoreFiles: oc.IgnoreFiles,
-		IgnorePaths: rc.IgnorePaths,
 	}
 	mc = mergeInRepoConfig(mc, orc, repo)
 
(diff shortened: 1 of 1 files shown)


jeffmendoza:
So "IgnorePaths" only exists in repo config, and "Ignore Files" only exists in orgconfig. This was a conscious design deicision. It looks like L 282 was supposed to "merge" the repo config.

I think the bug was a result of the "org-repo-config" feature, the `orc` variable in `mergeConfig()`.

Looking at it now, I think L282 should be removed, and given the patched lines you have added `orc` can be added with L284 and repoconfig with L287.

coheigea:
@jeffmendoza Thanks, I updated the PR if I understood you correctly.

jeffmendoza (approved):

jeffmendoza:
Looks great. Thanks for finding this! I'll let you know when it is rolled out.

coheigea:
Thanks a lot @jeffmendoza . We were wondering actually about when a new release might be, given that the last release was last March? Is there a reason we couldn't have monthly releases say?

jeffmendoza:
I'd be happy to do a release now. I've been meaning to go through the PRs and update whats-new.md

For monthly: no reason not to, we just need to see what we can use as a reminder/scheduling system to keep on top of them.

jeffmendoza:
This is deployed to prod now.

Issue #312, opened 2022-11-10
[Feature Request] Require review from Code Owners


testworksau:
It would be great if this tool could handle the setting of the Require Code Owner Reviews sub-property of branch protections. 

This is governed by the `require_code_owner_reviews` property in the `required_pull_request_reviews` object.

This might go hand in hand with https://github.com/ossf/allstar/issues/284 (or negate the need for it).

https://docs.github.com/en/rest/branches/branch-protection#update-branch-protection

We are happy to contribute if you're happy to see this added.

jeffmendoza:
Same as the other issue, happy to have more coverage of the BP options.

coheigea:
It can be closed I think @jeffmendoza  ?

PR #343, opened 2023-02-08
Add a config switch to require code owner reviews

coheigea:
Signed-off-by: Colm O hEigeartaigh <coheigea@apache.org>

Implemented https://github.com/ossf/allstar/issues/312

### DIFF
--- a/pkg/policies/branch/branch.go
+++ b/pkg/policies/branch/branch.go
@@ -52,6 +52,10 @@ type OrgConfig struct {
 	// RequireApproval : set to true to enforce approval on PRs, default true.
 	RequireApproval bool `json:"requireApproval"`
 
+	// RequireCodeOwnerReviews : set to true to enforce code owner reviews on PRs, default false.
+	// If set to true, then "requireApproval" must also be true
+	RequireCodeOwnerReviews bool `json:"requireCodeOwnerReviews"`
+
 	// ApprovalCount is the number of required PR approvals, default 1.
 	ApprovalCount int `json:"approvalCount"`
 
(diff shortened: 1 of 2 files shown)


pkg/policies/branch/branch.go
```diff
@@ -566,6 +582,10 @@ func fix(ctx context.Context, rep repositories, c *github.Client,
 				pr.RequiredPullRequestReviews.RequiredApprovingReviewCount = mc.ApprovalCount
 				update = true
 			}
+			if mc.RequireCodeOwnerReviews && !pr.RequiredPullRequestReviews.RequireCodeOwnerReviews {
```
jeffmendoza, line 586: Should L576 above also change to `if mc.RequireApproval || mc.RequireCodeOwnerReviews `?

jeffmendoza: Also L567

jeffmendoza (changes requested):
Looks great, thanks. A couple questions about the if statements.
Alternatively, The help for the config (L55) could stipulate that that `requireCodeOwnerReviews` only applies if `requireApproval` is true. This might simplify the logic a bit.

pkg/policies/branch/branch.go
coheigea: I've updated the help, as I think it doesn't make sense to use requireCodeOwnerReviews without requireApproval

jeffmendoza (approved):

jeffmendoza:
Thank you!

I'll do a little manual testing in staging before pushing to prod.
Do you have anything else in the works? I can do a release tomorrow if not.

coheigea:
No nothing else planned, thanks.

PR #355, opened 2023-02-15
Skip policy evaluation if the repo is not enabled

coheigea:
The allstar run takes a very long time on a github org which has many repos that are excluded by virtue of being archived/forked etc, with these repos being excluded via "optOut" config. Instead this PR skips policy evaluation when the repositories are not enabled. 

@jeffmendoza If you feel the old behaviour should be preserved, let me know and I'll make it configurable. I can't imagine most use-cases would want policy evaluation on repos you are opting out.

### DIFF
--- a/pkg/enforce/enforce.go
+++ b/pkg/enforce/enforce.go
@@ -23,6 +23,7 @@ import (
 	"time"
 
 	"github.com/ossf/allstar/pkg/config"
+	"github.com/ossf/allstar/pkg/config/operator"
 	"github.com/ossf/allstar/pkg/ghclients"
 	"github.com/ossf/allstar/pkg/issue"
 	"github.com/ossf/allstar/pkg/policies"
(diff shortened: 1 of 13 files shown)


jeffmendoza:
Hi @coheigea, this operator-level option is supposed to do this: https://github.com/ossf/allstar/blob/main/pkg/config/operator/operator.go#L58-L65

However, due to historical reasons, each policy is implementing it separately (ex: https://github.com/ossf/allstar/blob/main/pkg/policies/security/security.go#L120), and it is a bit messy and inconsistant.

I'd like to refactor that to be checked in the enforce loop as you have here. and remove the checks from each policy.

coheigea:
@jeffmendoza Thanks for the feedback, I made the changes along the lines you suggested - please take another look. 

I'm wondering though if moving the functionality into the enforcer means that it doesn't work any more if the repo is disabled at the policy level? Maybe I need to add a function to the policies to return if it's enabled or not.

jeffmendoza:
Yeah, I'm seeing that the logic is now changed - the `doNothingOnOptOut` option is hinged on the whole repo enable/disable and not individual policies.

I think this is better and more desirable behavior. At the `allstar.yaml` level you would have `optOutprivate/fork/archive` and then each individual policy would just have all enabled. That said, we should have this documented in the operator.go comment and in the changelog.

Yes, adding an `IsEnabled()` function to the policy definition makes sense, so the enforce loop can call that and then decide there before calling `Check`.

coheigea:
@jeffmendoza I added IsEnabled to the policies, please review again.

jeffmendoza (approved):

jeffmendoza:
Awesome!

Issue #358, opened 2023-02-28
Proposal: New policy for repo admins


coheigea:
I'd like to propose a new policy for repo admins (although I'm open to any other ideas) to be able to enforce some things we'd like to be able to monitor/control with allstar:

- bool ownerlessAllowed: OwnerlessAllowed defined if repositories are allowed to have no administrators, default false. This is copied from the outside collaborators policy, as IMO it belongs here as it's nothing to do with outside collaborators.
- bool userAdminsAllowed: Whether to allow users to be admins on a repo. If false then only teams can be admins. Default true.
- int maxNumberUserAdmins: Maximum number of user admins on a repo
- bool teamAdminsAllowed: Whether to allow teams to be admins on a repo. If false then only users can be admins. Default true.
- int maxNumberAdminTeams: Maximum number of teams with admin rights on a repo.

Also I'd add an exemption list to allow individual exemptions at the org level per-repo. Any feedback welcome.

jeffmendoza:
Sounds great. Maybe it would  good to propose the config structs/yamls for the new policy as well, and include the comment/documentation for each field to work through the logic.

I had included [`testingOwnerlessAllowed`](https://github.com/ossf/allstar/blob/main/pkg/policies/outside/outside.go#L80) in the "Outside Collaborator" policy, but a new separate policy makes sense. Because it is "testing" it would be ok to remove from that policy and include in a new admins policy.

coheigea:
It would be nice too have:

- int maxNumberMembersPerAdminTeam: The maximum number of members in a team with admin permissions on a repo that are allowed.

However, MembersCount returns 0 as per https://github.com/google/go-github/issues/1252.

PR #360, opened 2023-03-02
Moving TestingOwnerlessAllowed from outside collaborators policy to n…

coheigea:
…ew Admin policy

Part 1 of https://github.com/ossf/allstar/issues/358. I also improved the test coverage. Tested it locally against a github repo as well.

### DIFF
--- a/pkg/policies/admin/admin.go
+++ b/pkg/policies/admin/admin.go
@@ -1,4 +1,4 @@
-// Copyright 2021 Allstar Authors
+// Copyright 2023 Allstar Authors
 
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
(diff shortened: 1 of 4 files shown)


pkg/policies/admin/admin.go
```diff
@@ -0,0 +1,333 @@
...
+	OwnerlessAllowed bool
+	Exemptions       []*AdministratorExemption
+}
+
+type globCache map[string]glob.Glob
+
+// AdministratorExemption is an exemption entry for the Repository Administrators policy.
+type AdministratorExemption struct {
+
+	// Repo is a GitHub repo name
```
jeffmendoza, line 85: Should mention here that globs are allowed.

jeffmendoza:
Looks great. Just a comment on the documentation, since you are working on a follow up, I'll merge this and include that in the next one.

jeffmendoza (approved):

Issue #284, opened 2022-09-01
[Feature Request] Check for presence of CODEOWNERS file


emtunc:
Branch Protection is a good first step towards ensuring code is approved by _a_ member of the organisation. If we are to take this to the next level then we should be checking for a _valid_ CODEOWNERS file (presence of the file is sufficient to begin with).

I would like to see an issue raised when a CODEOWNERS file is missing in the root of a repository.

It's not a check that could have the "fix" action applied to it but that should be okay.

This could possibly sit under the _branch_protection.yaml_ config as you would typically only use CODEOWNERS as part of a Branch Protection stategy.

Gby56:
@jeffmendoza this can be closed

PR #380, opened 2023-03-21
WIP Feature: CODEOWNERS policy

Gby56:
Implements https://github.com/ossf/allstar/issues/284
I also had to bump `github.com/google/go-github` github REST client to v50, to access the function `GetCodeownersErrors` as it was not exposed in the API of the v43


Gby56:
Hello @jeffmendoza :)
I ran this locally to test things out, seems like it's good, I'm just not sure about what tests to implement yet
It's pretty simple, the API returns a 404 if no CODEOWNERS is found. The GraphQL API was a nightmare to figure out, so I had to use the REST one, and bump it (as explained above)

coheigea:
I'd recommend doing the version bump in a separate initial PR @Gby56 . Curious as to why dependabot didn't pick it up?

pkg/policies/codeowners/codeowners.go
```diff
@@ -0,0 +1,221 @@
...
+	log.Info().
+		Str("org", owner).
+		Str("repo", repo).
+		Str("area", polName).
+		Bool("enabled", enabled).
+		Msg("Check repo enabled")
+
+	codeownererrors, _, err := rep.GetCodeownersErrors(context.Background(), owner, repo)
+
+	if err != nil {
```
jeffmendoza, line 122: Probably want to differentiate here between an expected 404 and an unexpected error. See other code for examples: https://github.com/ossf/allstar/search?q=http.StatusNotFound&type=code

pkg/policies/codeowners/codeowners.go
```diff
@@ -0,0 +1,221 @@
...
+			Enabled:    enabled,
+			Pass:       true,
+			NotifyText: "",
+		}, nil
+	}
+
+	return &policydef.Result{
+		Enabled:    enabled,
+		Pass:       false,
+		NotifyText: "CODEOWNERS file not present.\n" + notifyText,
```
jeffmendoza, line 141: Wouldn't this case be that the file is present, but has errors?

pkg/policies/codeowners/codeowners.go
```diff
@@ -0,0 +1,221 @@
...
+}
+
+type mergedConfig struct {
+	Action string
+}
+
+type details struct {
+	Enabled bool
+	URL     string
+}
```
jeffmendoza, line 64: The details struct is what you would pass back as a part of the policydef.Result. You probably want a bool if it was found or not (404 or not), and a count of errors if found.

jeffmendoza (commented):
Looking good so far. I'll do another pass when the lint/test errors are fixed. Looks like some may be due to updating the github client verion.

jeffmendoza:
Thanks for bumping the github client version, dependabot doesn't because they usually bump the major version, and it can cause incompatabilites. A separate PR for that alone would be ok.

Gby56:
Thanks a lot ! I'll do a separate PR for the bump tomorrow then :) and yes you're right, I should count the codeowner errors and maybe still create an issue when it's broken, even if at least it exists (i think it's best)

pkg/enforce/enforce.go
```diff
@@ -130,6 +130,16 @@ func EnforceAll(ctx context.Context, ghc ghclients.GhClientsInterface) (EnforceA
...
+			// For debug usage, you can filter repositories to test against a single one
+			// TODO: Implement a cli flag option to specific a repository, and/or a policy you want to run, useful for testing.
+			// var found github.Repository
+			// for _, r := range repos {
+			// 	if *r.Name == "daikon-ee" {
+			// 		found = *r
+			// 	}
+			// }
+			// repos = []*github.Repository{}
+			// repos = append(repos, &found)
```
coheigea, line 142: Implement as new PR please :-)

pkg/policies/codeowners/codeowners.go
```diff
@@ -0,0 +1,221 @@
+// Copyright 2021 Allstar Authors
```
coheigea, line 1: 2023

pkg/policies/codeowners/codeowners_test.go
```diff
@@ -0,0 +1,259 @@
+// Copyright 2021 Allstar Authors
```
coheigea, line 1: 2023

pkg/policies/codeowners/codeowners.go
```diff
@@ -0,0 +1,221 @@
...
+	"github.com/rs/zerolog/log"
+)
+
+const configFile = "codeowners.yaml"
+const polName = "CODEOWNERS"
+
+const notifyText = `A CODEOWNERS file can give users information about who is responsible for the maintenance of the repository, or specific folders/files. This is different the access control/permissions on a repository.
+
+To fix this, add a CODEOWNERS file to your repository, following the official Github documentation and maybe your company's policy.
+https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/customizing-your-repository/about-code-owners`
```
coheigea, line 34: Put the backtick on a new line

pkg/policies/codeowners/codeowners.go
```diff
@@ -0,0 +1,221 @@
...
+// OrgConfig is the org-level config definition for Branch Protection.
+type OrgConfig struct {
+	// OptConfig is the standard org-level opt in/out config, RepoOverride applies to all
+	// BP config.
+	OptConfig config.OrgOptConfig `json:"optConfig"`
+
+	// Action defines which action to take, default log, other: issue...
+	Action string `json:"action"`
+
+	//TODO add default contents for "fix" action
```
coheigea, line 45: We should allow an exemption on a per-repo basis, so when specified we don't require Codeowners for the given repo. But this could be a separate PR if you want.

coheigea (changes requested):

PR #382, opened 2023-03-22
Bump google/go-github to v50

Gby56:
As we discussed, separated the v50 from my CODEOWNERS PR

PR #383, opened 2023-03-22
WIP: Implement CODEOWNERS detection policy

Gby56:
Depends on https://github.com/ossf/allstar/pull/382 for the version bump first

This will create issues if:
- A CODEOWNERS is not present (404)
- A CODEOWNERS is present but has errors (this is shown in the Github web UI but it doesn't create issues)

If you want an example of what it looks like:  https://github.com/Talend/tesb-studio-se/issues/1937


Gby56:
@jeffmendoza after the PR to bump https://github.com/ossf/allstar/pull/382 could you run checks here too ?
Thanks :)

Gby56:
Ok I messed up again I think, I merged my github client version bump into the branch, it created a conflict
I reverted, then tried to rebase my feature branch on the main (that I synced)... not sure what's going on with that diff...
image

PR #385, opened 2023-03-22
Implement --repo and --policy cli flags to filter down Allstar enforcement scope

Gby56:
Quite a useful feature imo, especially when you want to test Allstar locally

### DIFF
--- a/cmd/allstar/main.go
+++ b/cmd/allstar/main.go
@@ -52,8 +52,7 @@ func main() {
 		supportedPoliciesMap[policyName] = policyName
 		if i < len(supportedPolicies)-1 {
 			supportedPoliciesMsg += policyName + ", "
-		}
-		if i == len(supportedPolicies)-1 {
+		} else {
 			supportedPoliciesMsg += policyName
 		}
 	}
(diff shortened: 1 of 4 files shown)


cmd/allstar/main.go
```diff
@@ -41,12 +43,45 @@ func main() {
...
+	supportedPoliciesMap := map[string]string{}
+	var supportedPoliciesMsg = ""
 
+	for i, p := range supportedPolicies {
+		var policyName = p.Name()
+		supportedPoliciesMap[policyName] = policyName
+		if i < len(supportedPolicies)-1 {
+			supportedPoliciesMsg += policyName + ", "
+		}
+		if i == len(supportedPolicies)-1 {
```
coheigea: Can we just use else here?

coheigea (changes requested):
Great improvement for testing

cmd/allstar/main.go
Gby56: Done

coheigea (dismissed):

Gby56:
I need to fix the tests !

jeffmendoza:
Sorry they were not running automatically
image
I just changed the setting from middle to top here, hope that helps.

jeffmendoza (approved):

PR #387, opened 2023-03-23
Feature: Add Operator repo allowlist

### LINKED ISSUE #365: Feature: GitHub installation allowlist
## Description

Allstar should allow the Operator to configure an allowlist of repositories/organizations that it can be installed on. If it detects an installation of the GitHub app on a repository that's not in this allowlist, it should skip the repository or deactivate the installation.

## What is the current behavior?

Today, Allstar instances that are published on the GitHub app marketplace as public apps can be installed by anyone. To privately install Allstar (e.g. enterprise), the operator must create a private app installation per GitHub organization, which is time consuming for operators with lots of GitHub organizations to manage. This may lead to mixing of public repo contents and 'enterprise' repo contents the operator intends to use Allstar on.

raghavkaul:
Allow operators to configure a repo Allowlist. The repo allowlist accepts a list of comma-separated globs. 

# New Behavior
Repos that match any glob in the list can be scanned by Allstar. Repos that don't match any of the globs in the list are not scanned by Allstar and are removed from the installation.

# Notes for the reviewer
To make change management easier, if the operator hasn't configured an Allowlist, we fail open. This is so that people don't update Allstar and mistakenly uninstall all of their installations just because they were unaware of this feature. In future versions, we could make this parameter mandatory.

Fixes #365.

cc: @azeemshaikh38

### DIFF
--- a/pkg/config/operator/operator.go
+++ b/pkg/config/operator/operator.go
@@ -18,6 +18,7 @@ package operator
 import (
 	"os"
 	"strconv"
+	"strings"
 	"time"
 
 	"github.com/rs/zerolog"
@@ -83,6 +84,11 @@ const GitHubIssueFooter = `This issue will auto resolve when the policy is in co
 
 Issue created by Allstar. See https://github.com/ossf/allstar/ for more information. For questions specific to the repository, please contact the owner or maintainer.`
 
+// AllowedRepositories is the set of GitHub repositories on which this Allstar instance
+// is allowed to be installed. This allows a public GitHub app to be shared between GitHub
+// organizations and repos while restricting installation of the app
+var AllowedRepositories []string
+
 // NoticePingDuration is the duration (in hours) to wait between pinging notice actions,
 // such as updating a GitHub issue.
 const setNoticePingDurationHrs = (24 * time.Hour)
@@ -138,4 +144,7 @@ func setVars() {
 	} else {
 		NoticePingDuration = setNoticePingDurationHrs
 	}
+
+	allowedRepositories := os.Getenv("GITHUB_ALLOWED_REPOS")
+	AllowedRepositories = strings.Split(allowedRepositories, ",")
 }
--- a/pkg/config/operator/operator.go
+++ b/pkg/config/operator/operator.go
@@ -84,10 +84,10 @@ const GitHubIssueFooter = `This issue will auto resolve when the policy is in co
 
 Issue created by Allstar. See https://github.com/ossf/allstar/ for more information. For questions specific to the repository, please contact the owner or maintainer.`
 
-// AllowedRepositories is the set of GitHub repositories on which this Allstar instance
+// AllowedOrganizations is the set of GitHub repositories on which this Allstar instance
 // is allowed to be installed. This allows a public GitHub app to be shared between GitHub
 // organizations and repos while restricting installation of the app
-var AllowedRepositories []string
+var AllowedOrganizations []string
 
 // NoticePingDuration is the duration (in hours) to wait between pinging notice actions,
 // such as updating a GitHub issue.
@@ -145,6 +145,6 @@ func setVars() {
 		NoticePingDuration = setNoticePingDurationHrs
 	}
 
-	allowedRepositories := os.Getenv("GITHUB_ALLOWED_REPOS")
-	AllowedRepositories = strings.Split(allowedRepositories, ",")
+	allowedOrgs := os.Getenv("GITHUB_ALLOWED_ORGS")
+	AllowedOrganizations = strings.Split(allowedOrgs, ",")
 }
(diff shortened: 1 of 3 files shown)


jeffmendoza:
Any reason to make this a repo allowlist? #365 discussed an installation (org) allowlist, and the action would be to uninstall allstar. This is just removing specific repos from an installation. It might end up leaving installations with no repos, which would be weird. GitHub might even error, trying to remove the last one (haven't tested).

pkg/config/operator/operator.go
```diff
@@ -83,6 +84,11 @@ const GitHubIssueFooter = `This issue will auto resolve when the policy is in co
 
 Issue created by Allstar. See https://github.com/ossf/allstar/ for more information. For questions specific to the repository, please contact the owner or maintainer.`
 
+// AllowedRepositories is the set of GitHub repositories on which this Allstar instance
+// is allowed to be installed. This allows a public GitHub app to be shared between GitHub
+// organizations and repos while restricting installation of the app
+var AllowedRepositories []string
```
azeemshaikh38: Should be allowedOrgs instead of repos?

pkg/enforce/enforce.go
```diff
@@ -106,7 +111,7 @@ func EnforceAll(ctx context.Context, ghc ghclients.GhClientsInterface) (EnforceA
 
 		g.Go(func() error {
 
-			repos, _, err := getAppInstallationRepos(ctx, ic)
+			repos, _, err := getAppInstallationRepos(ctx, ic, iid)
```
azeemshaikh38: This check should be before the gothread within the for loop. Check if the installation is included within the allowed_org list and uninstall otherwise

raghavkaul:
@jeffmendoza Yes, that makes sense. Changed it to an organization allowlist. It was initially a repo allowlist with globs because that felt more flexible, and only disabled the installation because that felt less destructive/less likely to error.

pkg/enforce/enforce.go
```diff
@@ -204,25 +208,72 @@ func runPoliciesOnInstRepos(ctx context.Context, repos []*github.Repository, ghc
...
 		PerPage: 100,
 	}
+	var err error
 	for {
-		is, resp, err := ac.Apps.ListInstallations(ctx, opt)
+		is, resp, err := ac.Apps.ListInstallations(ctx, opts)
+
 		if err != nil {
-			return nil, err
+			return insts, err
```
jeffmendoza: `return nil, err`

pkg/enforce/enforce.go
```diff
@@ -204,25 +208,72 @@ func runPoliciesOnInstRepos(ctx context.Context, repos []*github.Repository, ghc
...
+	if err != nil {
+		return nil, err
 	}
+
+	if len(operator.AllowedOrganizations) == 0 {
+		return is, nil
+	}
+
+	var insts []*github.Installation
+	for i := range is {
```
jeffmendoza: `for _, i:= range is {`

pkg/enforce/enforce.go
```diff
@@ -204,25 +208,72 @@ func runPoliciesOnInstRepos(ctx context.Context, repos []*github.Repository, ghc
...
+		var m bool
+		for j := range operator.AllowedOrganizations {
```
jeffmendoza: `for _, ao := range operator.AllowedOrganizations {`

pkg/enforce/enforce.go
```diff
@@ -204,25 +208,72 @@ func runPoliciesOnInstRepos(ctx context.Context, repos []*github.Repository, ghc
...
+			// Account.Login holds the name of the GitHub Organization
+			// that Allstar is installed on
+			if operator.AllowedOrganizations[j] == *is[i].Account.Login {
```
jeffmendoza: `if i.Account != nil && ao == i.Account.GetLogin() {`

pkg/enforce/enforce.go
```diff
@@ -204,25 +208,72 @@ func runPoliciesOnInstRepos(ctx context.Context, repos []*github.Repository, ghc
...
+				m = true
+			}
+		}
+		if !m {
+			resp, err := deleteInstallation(ctx, ac, *is[i].ID)
```
jeffmendoza: `resp, err := deleteInstallation(ctx, ac, i.GetID())`

pkg/enforce/enforce.go
```diff
@@ -204,25 +208,72 @@ func runPoliciesOnInstRepos(ctx context.Context, repos []*github.Repository, ghc
...
+			if err != nil || resp.StatusCode != 200 {
+				log.Error().
+					Err(err).
```
jeffmendoza, line 260: Also log:
```
				Str("area", "bot").
				Str("account", i.Account.GetLogin()).
  				Int64("id", i.GetID()).
```

jeffmendoza (changes requested):

jeffmendoza (dismissed):

jeffmendoza:
Looks good, need to resolve conflicts.

jeffmendoza (approved):

PR #388, opened 2023-03-27
Implement CODEOWNERS policy, with tests

Gby56:
Sorry about the failed PR attempts before @jeffmendoza 😅 
Finally got the hang of Golang tests, was a bit tricky to implement the deep nested Github errors but it works !
Implements https://github.com/ossf/allstar/issues/284

### DIFF
--- a/pkg/policies/codeowners/codeowners.go
+++ b/pkg/policies/codeowners/codeowners.go
@@ -18,6 +18,7 @@ package codeowners
 import (
 	"context"
 	"fmt"
+	"net/http"
 
 	"github.com/ossf/allstar/pkg/config"
 	"github.com/ossf/allstar/pkg/policydef"
@@ -64,10 +65,9 @@ type mergedConfig struct {
 }
 
 type details struct {
-	Enabled         bool
-	URL             string
-	CodeownersFound bool
-	ErrorCount      int
+	CodeownersFound  bool
+	ErrorCount       int
+	CodeownersErrors github.CodeownersErrors
 }
 
 var configFetchConfig func(context.Context, *github.Client, string, string, string, config.ConfigLevel, interface{}) error
@@ -110,12 +110,13 @@ func check(ctx context.Context, rep repositories, c *github.Client, owner,
 	repo string) (*policydef.Result, error) {
 	oc, orc, rc := getConfig(ctx, c, owner, repo)
 	enabled, err := configIsEnabled(ctx, oc.OptConfig, orc.OptConfig, rc.OptConfig, c, owner, repo)
-	var d details
-	d.Enabled = enabled
-	d.URL = ""
+
 	if err != nil {
 		return nil, err
 	}
+
+	var d details
+
 	log.Info().
 		Str("org", owner).
 		Str("repo", repo).
@@ -123,10 +124,36 @@ func check(ctx context.Context, rep repositories, c *github.Client, owner,
 		Bool("enabled", enabled).
 		Msg("Check repo enabled")
 
-	codeownererrors, _, err := rep.GetCodeownersErrors(context.Background(), owner, repo)
+	codeownererrors, resp, err := rep.GetCodeownersErrors(ctx, owner, repo)
 
-	// this is a 404, CODEOWNERS is not the in the repository
-	if err != nil {
+	if err == nil {
+		// "CODEOWNERS" exists
+		d.ErrorCount = len(codeownererrors.Errors)
+		d.CodeownersFound = true
+		// the CODEOWNERS is present and has no errors, pass
+		if d.ErrorCount == 0 {
+			return &policydef.Result{
+				Enabled:    enabled,
+				Pass:       true,
+				NotifyText: "",
+				Details:    d,
+			}, nil
+		}
+		// otherwise, fail because CODEOWNERS exists and has errors
+		d.CodeownersErrors = *codeownererrors
+		var errorMessage = fmt.Sprintf("%s\nCODEOWNERS file present but has %d errors.\n", notifyText, d.ErrorCount)
+		for _, e := range codeownererrors.Errors {
+			errorMessage += fmt.Sprintf("- %s\n  - %s\n", e.Path, e.Message)
+		}
+		return &policydef.Result{
+			Enabled:    enabled,
+			Pass:       false,
+			NotifyText: errorMessage,
+			Details:    d,
+		}, nil
+
+	} else if resp != nil && resp.StatusCode == http.StatusNotFound {
+		// "CODEOWNERS" does not exist, err is also not nil but we don't need it
 		d.CodeownersFound = false
 		return &policydef.Result{
 			Enabled:    enabled,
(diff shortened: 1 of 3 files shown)


pkg/policies/codeowners/codeowners.go
```diff
@@ -0,0 +1,240 @@
...
+}
+
+func check(ctx context.Context, rep repositories, c *github.Client, owner,
+	repo string) (*policydef.Result, error) {
+	oc, orc, rc := getConfig(ctx, c, owner, repo)
+	enabled, err := configIsEnabled(ctx, oc.OptConfig, orc.OptConfig, rc.OptConfig, c, owner, repo)
+	var d details
+	d.Enabled = enabled
+	d.URL = ""
+	if err != nil {
```
jeffmendoza, line 121: Put the error check right after the call.

pkg/policies/codeowners/codeowners.go
```diff
@@ -0,0 +1,240 @@
...
+	return configIsEnabled(ctx, oc.OptConfig, orc.OptConfig, rc.OptConfig, c, owner, repo)
+}
+
+func check(ctx context.Context, rep repositories, c *github.Client, owner,
+	repo string) (*policydef.Result, error) {
+	oc, orc, rc := getConfig(ctx, c, owner, repo)
+	enabled, err := configIsEnabled(ctx, oc.OptConfig, orc.OptConfig, rc.OptConfig, c, owner, repo)
+	var d details
+	d.Enabled = enabled
+	d.URL = ""
```
jeffmendoza: Nil values don't need to be initialized.
```
d := details{Enbaled: enabled}
```

pkg/policies/codeowners/codeowners.go
```diff
@@ -0,0 +1,240 @@
...
+		return nil, err
+	}
+	log.Info().
+		Str("org", owner).
+		Str("repo", repo).
+		Str("area", polName).
+		Bool("enabled", enabled).
+		Msg("Check repo enabled")
+
+	codeownererrors, _, err := rep.GetCodeownersErrors(context.Background(), owner, repo)
```
jeffmendoza: use the passed-in `ctx`.

pkg/policies/codeowners/codeowners.go
```diff
@@ -0,0 +1,240 @@
...
+
+	// this is a 404, CODEOWNERS is not the in the repository
+	if err != nil {
```
jeffmendoza: You need to check the response here. It could be a 404, or another error
example: https://github.com/ossf/allstar/blob/9fd9b106f0cd1f3754c461d9c123d39a89024f6e/pkg/config/location.go#L91-L108

pkg/policies/codeowners/codeowners.go
```diff
@@ -0,0 +1,240 @@
...
+			Enabled:    enabled,
+			Pass:       true,
+			NotifyText: "",
+			Details:    d,
+		}, nil
+	}
+	// otherwise, fail because CODEOWNERS has errors
+	var errorMessage = fmt.Sprintf("CODEOWNERS file present but has %d errors.\n", d.ErrorCount)
+	for _, e := range codeownererrors.Errors {
+		errorMessage += fmt.Sprintf("- %s\n  - %s", e.Path, e.Message)
```
jeffmendoza: Go ahead and add these to the details as well.

pkg/policies/codeowners/codeowners.go
jeffmendoza: That said, looking through the function, `URL` is not used, and `Enabled` is a part of `policydef.Result`, so it doesn't need to be in details.

pkg/policies/codeowners/codeowners.go
jeffmendoza: Probably need a `\n` at end of the message? `"- %s  - %s\n"`

pkg/policies/codeowners/codeowners.go
```diff
@@ -0,0 +1,240 @@
...
+	if d.ErrorCount == 0 {
+		return &policydef.Result{
+			Enabled:    enabled,
+			Pass:       true,
+			NotifyText: "",
+			Details:    d,
+		}, nil
+	}
+	// otherwise, fail because CODEOWNERS has errors
+	var errorMessage = fmt.Sprintf("CODEOWNERS file present but has %d errors.\n", d.ErrorCount)
```
jeffmendoza: Probably want to start this with `notifyText`?

jeffmendoza: or add to the end

pkg/policies/codeowners/codeowners.go
```diff
@@ -0,0 +1,240 @@
...
+type mergedConfig struct {
+	Action string
+}
+
+type details struct {
+	Enabled         bool
+	URL             string
+	CodeownersFound bool
+	ErrorCount      int
+}
```
jeffmendoza, line 78: Looking through below, looks like details should be something like:

```
type details struct {
	CodeownersFound bool
	ErrorCount int
	Errors []CodeOwnerError
}
type CodeOwnerError struct {
	Message string
	Path string
}
```

jeffmendoza (changes requested):

jeffmendoza:
> Sorry about the failed PR attempts before

No worries! feel free to ask about anything if you are stuck.

pkg/policies/codeowners/codeowners.go
Gby56: Yeah it's a weird pattern, from what I can see when debugging, both the response and the `err` have the 404 information, because the error has the `github.Response` inside of it
Maybe I could handle different cases like a 500, and not create a Github issue at all in that case

pkg/policies/codeowners/codeowners.go
Gby56: I just added a simple field `CodeownersErrors github.CodeownersErrors` in the details struct, so all the fields are in there

Gby56:
I started going into the rabbithole of "what if it's not a 404 ?" and writing test and all, but I just chose to ignore and return nil instead of the error, Allstar won't take any action on HTTP 500 responses I guess anyways

Gby56:
Added a `requireCODEOWNERS` config option for both Org and Repo-level, so that by default, set to false, the Policy won't create issues for repos that don't have a CODEOWNERS.
It will only create issues for repos that have one, but contains errors (makes it easier to alert people at least, even if Github shows you a nice banner when there are errors detected...)
If you want to create issues on repos if they don't have a CODEOWNERS, just set `requireCODEOWNERS: true` on the org or repo config

Tested locally, switched on, and then off, the `requireCODEOWNERS` option, it created an issue on the repo, then closed it.

pkg/policies/codeowners/codeowners.go
```diff
@@ -0,0 +1,271 @@
...
+			Pass:       true,
+			NotifyText: "CODEOWNERS file not present.\n" + notifyText,
+			Details:    d,
+		}, nil
+
+	} else {
+		// Unknown error getting "CODEOWNERS", this could be an HTTP 500
+		return nil, nil
+	}
+
```
jeffmendoza: `return nil, err`
Policies propagate the errors back up the stack if it is unexpected.
Also, you can remove the `else { }` and just have the return at the end of the function.

jeffmendoza (approved):

jeffmendoza:
Awesome!

Gby56:
https://github.com/ossf/allstar/issues/284 This should be closed then :)

PR #416, opened 2023-05-19
Updates on CI jobs

cpanato:
- fix dependabot config
- use ko action to install ko
- upgrade to cosign v2

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -8,4 +8,4 @@ updates:
 - package-ecosystem: "github-actions"
   directory: "/"
   schedule:
-      interval: "daily"
+    interval: "daily"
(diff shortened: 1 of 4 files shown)


cpanato:
cc @justaugustus

jeffmendoza:
I want to keep the read-only workflows using major version tag aliases for actions
Yes! I do want to update cosign, thanks for that.

cpanato:
> I want to keep the read-only workflows using major version tag aliases for actions Yes! I do want to update cosign, thanks for that.

major version tags are mutating and we don't know actually what is the actual release, also dependabot can deal with that update, and be will easier for us, but it is your call, want me to drop those things? did that because this is one of the practices that we in ossf advocate for :)

jeffmendoza:
Yep, for read-only workflows it is easier to just use the mutable tag.

cpanato:
@jeffmendoza reverted

jeffmendoza (approved):

PR #418, opened 2023-05-22
refactor release job and sign the image digest and not the tag

### LINKED ISSUE #417: Sign correct container when publishing releases
In the release workflow, we sign the container that is publicized for the release, which uses the version tag of the release:

https://github.com/ossf/allstar/blob/main/.github/workflows/release.yaml#L25

I was reading that it is not best practice to sign containers with mutable tags, and to only sign the digest tag.

cpanato:
- refactor release job and sign the image digest and not the tag

make the job a bit clear to read, add version comments and use the ko flags to output the image reference and the tag flag to push the image with the tag version to avoid the docker pull/tag/repush commands

Fixes: https://github.com/ossf/allstar/issues/417

/assign @jeffmendoza

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -1,28 +1,40 @@
+name: release
+
 on:
   push:
     tags:
       - '*'
+
 permissions:
   id-token: write # Undocumented OIDC support.
   packages: write # To publish container images to GHCR
   contents: write # To create a release
+
 jobs:
   release:
     runs-on: ubuntu-20.04
     steps:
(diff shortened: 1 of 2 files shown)


cpanato:
rehearsal:
https://github.com/cpanato/allstar/actions/runs/5047061522/jobs/9053399550
https://github.com/cpanato/allstar/releases/tag/v99.99.00

jeffmendoza (approved):

jeffmendoza:
Great! thanks!

PR #423, opened 2023-05-30
Always set ApprovalCount to 0 when RequireApproval is false

### LINKED ISSUE #419: Not requiring approvals does not imply approval count of 0?
Consider this either a question or a bug report 😄 

When setting
```
requireApproval: false
```
in `branch_protection.yaml`, one still gets the error
> PR Approvals below threshold 0 : 1 for branch main

if "Require a pull request before merging" is active and "Require approvals" is not active in a repository.

Question: Shouldn't `requireApproval: false` imply `approvalCount: 0`? And if not: Should this "surprising" (at least for me) behaviour be documented?

jamestran201:
Fixes https://github.com/ossf/allstar/issues/419

This PR explicitly sets the `ApprovalCount` to 0 when `RequireApproval` is `false` because not requiring approval implies that zero approval count is needed.

### DIFF
--- a/pkg/policies/branch/branch.go
+++ b/pkg/policies/branch/branch.go
@@ -49,6 +49,7 @@ type OrgConfig struct {
 	EnforceBranches map[string][]string `json:"enforceBranches"`
 
 	// RequireApproval : set to true to enforce approval on PRs, default true.
+	// When this config is false, ApprovalCount will always be set to 0.
 	RequireApproval bool `json:"requireApproval"`
 
 	// RequireCodeOwnerReviews : set to true to enforce code owner reviews on PRs, default false.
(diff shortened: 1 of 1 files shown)


pkg/policies/branch/branch.go
```diff
@@ -758,6 +759,11 @@ func mergeConfig(oc *OrgConfig, orc, rc *RepoConfig, repo string) *mergedConfig
 	if !oc.OptConfig.DisableRepoOverride {
 		mc = mergeInRepoConfig(mc, rc, repo)
 	}
+
+	if !mc.RequireApproval {
+		mc.ApprovalCount = 0
+	}
```
jamestran201, line 765: Please let me know if this is not the right place to add the change, thanks!

jeffmendoza (approved):

jeffmendoza:
Thanks!

Issue #422, opened 2023-05-25
Scorecards returning empty findings


coheigea:
With the latest code, we are seeing issues created from scorecards like:
```
Rule Description
This is a generic passthrough policy that runs the configured checks from Security Scorecards. Please see the [Security Scorecards Documentation](https://github.com/ossf/scorecard/blob/main/docs/checks.md) for more information on each check.

Results from policy: Token-Permissions : non read-only tokens detected in GitHub workflows

[0]:
[0]:
[0]:
[0]:
[0]:
[0]:
[0]:
```

PR #424, opened 2023-05-31
Scorecards returning empty findings

coheigea:
A fix for https://github.com/ossf/allstar/issues/422

With this commit https://github.com/ossf/scorecard/commit/2ea140a3ee9279904f29ed335b2bbef0923f242e, it seems the msg.X values aren't populated any more for token permissions. Instead we can get the information from l.Msg.Finding.

Tested it locally.

### DIFF
--- a/pkg/policies/scorecard/scorecard.go
+++ b/pkg/policies/scorecard/scorecard.go
@@ -225,7 +225,11 @@ func listJoin(list []string) string {
 func convertLogs(logs []checker.CheckDetail) []string {
 	var s []string
 	for _, l := range logs {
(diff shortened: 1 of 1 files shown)


coheigea:
@jeffmendoza Please take a look

jeffmendoza (approved):

jeffmendoza:
Sorry, missed this earlier.

spencerschrock:
This was not intentional from the Scorecard side. While we're working on a new results format, we've been trying to avoid changing the current one by gating things behind a feature flag. This slipped through. @laurentsimon FYI.

laurentsimon:
Good catch. That's why I'd like more e2e tests for SARIF https://github.com/ossf/scorecard/issues/2665. /cc @naveensrinivasan

PR #438, opened 2023-07-28
Feature Request: Use chainguard/busybox as the base image to support use as a GitHub Action

pauldoomgov:
We would like the option of running AllStar as a GitHub Action. The current container image uses `cgr.dev/chainguard/static` which is an excellent minimal base with very little surface area. Unfortunately, GitHub Actions requires `tail` to be available for use as a container:

~~~sh
/usr/bin/docker create --name ... --label ... --workdir /__w/.allstar/.allstar --network ...  -e "HOME=/github/home" -e GITHUB_ACTIONS=true -e CI=true -v "/var/run/docker.sock":"/var/run/docker.sock" -v "/home/runner/work":"/__w" -v "/home/runner/runners/2.306.0/externals":"/__e":ro -v "/home/runner/work/_temp":"/__w/_temp" -v "/home/runner/work/_actions":"/__w/_actions" -v "/opt/hostedtoolcache":"/__t" -v "/home/runner/work/_temp/_github_home":"/github/home" -v "/home/runner/work/_temp/_github_workflow":"/github/workflow" --entrypoint "tail" ghcr.io/ossf/allstar:v3.0 "-f" "/dev/null"
~~~

This PR switches the base image to `cgr.dev/chainguard/busybox` which adds `busybox` on top of `static`. This allong with the `-once` flag makes it possible to run AllStar in GitHub Actions.

If you can not accept this PR due to the additional surface area an alternate tack would be to build a second image for users who wish to run `allstar` as an action, while keeping the original pristine image as the default.

### DIFF
--- a/pkg/policies/scorecard/scorecard.go
+++ b/pkg/policies/scorecard/scorecard.go
@@ -189,8 +189,9 @@ func (b Scorecard) Check(ctx context.Context, c *github.Client, owner,
 
 **Rule Description**
 This is a generic passthrough policy that runs the configured checks from Security Scorecards. Please see the [Security Scorecards Documentation](https://github.com/ossf/scorecard/blob/main/docs/checks.md) for more information on each check.
-
+The score was %v, and the passing threshold is %v.
 `
+				notify = fmt.Sprintf(notify, res.Score, mc.Threshold)
 			}
 			if len(logs) > 10 {
 				notify += fmt.Sprintf(


jeffmendoza:
Hi Paul, happy to add this. I'm thinking a separate image called "allstar-busybox" would be great for those that need it, and keep the "allstar" static-binary-only image for those that don't. How does that sound?

Also, remember to commit with "git -s" for DCO (https://wiki.linuxfoundation.org/dco), Thanks!

pauldoomgov:
> Hi Paul, happy to add this. I'm thinking a separate image called "allstar-busybox" would be great for those that need it, and keep the "allstar" static-binary-only image for those that don't. How does that sound?
> 
> Also, remember to commit with "git -s" for DCO (https://wiki.linuxfoundation.org/dco), Thanks!

Excellent! I will update the PR to add the second image and make sure to include a DCO sign off.

PR #471, opened 2023-11-06
Skip empty repositories for Allstar policy enforcement

raghavkaul:
Skip empty repositories for Allstar policy enforcement. Check whether they're empty using [GitHub REST API](https://docs.github.com/en/rest/metrics/statistics?apiVersion=2022-11-28#get-all-contributor-commit-activity).

### DIFF
--- a/pkg/enforce/enforce.go
+++ b/pkg/enforce/enforce.go
@@ -19,6 +19,7 @@ package enforce
 import (
 	"context"
 	"fmt"
+	"net/http"
 	"sync"
 	"time"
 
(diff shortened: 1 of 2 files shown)


jeffmendoza:
Looking at https://docs.github.com/en/rest/metrics/statistics?apiVersion=2022-11-28#best-practices-for-caching and https://pkg.go.dev/github.com/google/go-github/v56/github#RepositoriesService.ListContributorsStats This doesn't seem like a good option to use. It will likely result in an error/202 and we don't need this expensive operation to tell if the repo is empty.

Is using something like https://pkg.go.dev/github.com/google/go-github/v56/github#RepositoriesService.GetContents an option and seeing if the top-level is empty?

raghavkaul:
Yes, that API makes sense. Updated!

pkg/enforce/enforce.go
```diff
@@ -318,6 +321,24 @@ func EnforceJob(ctx context.Context, ghc *ghclients.GHClients, d time.Duration,
...
+	}
+	if file == nil && len(dirs) == 0 {
+		log.Info().
+			Str("org", owner).
+			Str("repo", repo).
+			Str("area", policy).
+			Msg("Repository is empty, skipping")
+	}
+
+	return true, nil
```
jeffmendoza: Probably want to return false here, and true inside the if statement.

raghavkaul: Essential catch!

jeffmendoza (approved):

PR #483, opened 2024-01-08
Don't create issues for dangerous workflows when we have an inconclus…

coheigea:
…ive result

Since the recent update we are getting a ton of issues created for dangerous workflows, when the repo in question don't contain any github action workflows. The PR changes the check to pass if we have an inconclusive result (no workflows found)


coheigea:
FAO @jeffmendoza

jeffmendoza (approved):

jeffmendoza:
Do you think this is a bug in Scorecard, or just a new result that we were not processing correctly here in this Allstar policy?

coheigea:
I don't think it's a bug in scorecard, we just weren't properly checking for an inconclusive result.

PR #497, opened 2024-03-07
Switch to using a single worker

jeffmendoza:
Change "workers" cli option to be in pkg/config/operator and use ALLSTAR_NUM_WORKERS envvar with same default at 5. Update staging and prod config to use 1 worker to save concurrent memory usage.

Also, make sure to break out of enforceAll loop when context is cancelled.


jeffmendoza:
cc @raghavkaul I'm looking to use this in prod, and I don't have a good way to use cli options now, jut env vars. The initial setup was for all operator config to be in pkg/config/operator. Some cli options were added later for test/debug purposes.

Issue #501, opened 2024-03-19
[bug] Issue in member repo is updated/edited when org config IssueRepo is set centrally


twelsh-aw:
Noticed that there is a bug updating issues when `IssueRepo` [config setting](https://github.com/ossf/allstar/blob/964a34c8c454d9d3605fcd482c5cc2752ea4545b/pkg/config/config.go#L55) is set.

The bug is [here](https://github.com/ossf/allstar/blob/964a34c8c454d9d3605fcd482c5cc2752ea4545b/pkg/issue/issue.go#L142) and [here](https://github.com/ossf/allstar/blob/964a34c8c454d9d3605fcd482c5cc2752ea4545b/pkg/issue/issue.go#L162). Notice that `issueRepo` var is not used on these API calls like the other cases and `repo` is instead.

We are operating Allstar ourselves as a GHA on cron, so might not see logs on your end, but and example of this is exists in a private repo from our org

I am happy to submit a patch

jeffmendoza:
Thanks for fixing this!!

PR #502, opened 2024-03-20
Update issue in IssueRepo when change detected

twelsh-aw:
This was trying (and depending on app permissions, succeeding) at changing issue descriptions in repos directly even when IssueRepo was set. We update to obey IssueRepo config setting in this case


twelsh-aw:
fixes #501

jeffmendoza (approved):

Issue #535, opened 2024-07-02
Update emitted copy around OpenSSF Scorecard checks


justaugustus:
Using https://github.com/uwu-tools/ggreconcile/issues/8 as an example:

> _This issue was automatically created by [Allstar](https://github.com/ossf/allstar/)._
> 
> **Security Policy Violation** Project is out of compliance with Security Scorecards policy
> 
> **Rule Description** This is a generic passthrough policy that runs the configured checks from Security Scorecards. Please see the [Security Scorecards Documentation](https://github.com/ossf/scorecard/blob/main/docs/checks.md) for more information on each check. The score was 0, and the passing threshold is 10. **Results from policy: Branch-Protection : branch protection not enabled on development/release branches**
> 
>     * [0]:branch protection not enabled for branch 'main'
> 
> 
> **Results from policy: CI-Tests : 0 out of 2 merged PRs checked by a CI test -- score normalized to 0**
> 
>     * [0]:merged PR 4 without CI test at HEAD: [749a733](https://github.com/uwu-tools/ggreconcile/commit/749a733844dcab5f3b6a9f9acb3a5862face33c0)
> 
>     * [0]:merged PR 2 without CI test at HEAD: [2464e9c](https://github.com/uwu-tools/ggreconcile/commit/2464e9c0236dc7f3738e22eb218b0c61f5e87b3e)
> 
> 
> ⚠️ There is an updated version of this policy result! [Click here to see the latest update](https://github.com/uwu-tools/ggreconcile/issues/8#issuecomment-2204288161)
> 
> This issue will auto resolve when the policy is in compliance.
> 
> Issue created by Allstar. See https://github.com/ossf/allstar/ for more information. For questions specific to the repository, please contact the owner or maintainer.

There are some things that likely need to be updated within the check's copy e.g., "Security Scorecards" should be "OpenSSF Scorecard" (ref: https://github.com/ossf/scorecard/issues/2427, https://github.com/ossf/scorecard/pull/2428)

PR #536, opened 2024-07-02
Correct references to OpenSSF Scorecard

justaugustus:
Partially addresses https://github.com/ossf/allstar/issues/535
(it may fully address the issue, but for now, I'd like to leave that one open for additional investigation)

- docs: Correct instances of "Security Scorecards" to "OpenSSF Scorecard"

  Includes minor fixes to surrounding documentation.

- README: Correct Scorecard API URL

### DIFF
--- a/pkg/policies/scorecard/scorecard.go
+++ b/pkg/policies/scorecard/scorecard.go
@@ -12,7 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-// Package scorecard implements the generic Security Scorecards policy
+// Package scorecard implements the generic OpenSSF Scorecard policy
 package scorecard
 
 import (
@@ -31,7 +31,7 @@ import (
 )
 
 const configFile = "scorecard.yaml"
-const polName = "Security Scorecards"
+const polName = "OpenSSF Scorecard"
 
 // OrgConfig is the org-level config definition for this policy.
 type OrgConfig struct {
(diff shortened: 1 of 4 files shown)


pkg/policies/scorecard/scorecard.go
```diff
@@ -31,7 +31,7 @@ import (
 )
 
 const configFile = "scorecard.yaml"
-const polName = "Security Scorecards"
+const polName = "OpenSSF Scorecard"
```
justaugustus, line 34: @jeffmendoza — is this breaking? It's local, so I'm guessing "no", but just wanted to flag it on the review.

jeffmendoza: So this will change the name needed for the `-policy` cli option to only run this policy. https://github.com/ossf/allstar/blob/main/cmd/allstar/main.go#L46-L62

It will change the "area" value for all the structured logging in this policy.

It will change the name of the policy in the structured summary results that allstar logs on each loop https://github.com/ossf/allstar/blob/main/pkg/enforce/enforce.go#L157-L160

I think this is ok and is just something to call out in release notes.

jeffmendoza (approved):

pkg/policies/scorecard/scorecard.go
justaugustus: Thanks for the pointers, @jeffmendoza!
I've filed https://github.com/ossf/allstar/issues/537 to track this.

PR #570, opened 2024-09-24
Support globs for optOut/optInRepos

coheigea:
Just a draft solution with no tests thus far. @jeffmendoza what do you think of the general approach? I find it hard to use optOutRepos without having a glob solution.


jeffmendoza:
Awesome, this sounds great.

Will definitely need to update the docs: https://github.com/ossf/allstar/blob/main/README.md?plain=1#L91
The `contains` function should probably be renamed.
Yes, a test would be good.

coheigea:
Thanks @jeffmendoza , it's ready for review now

jeffmendoza (approved):

jeffmendoza:
Awesome, thanks!

PR #582, opened 2024-10-08
Update scorecard interface to use scorecard.Run()

jeffmendoza:
The "continue on error" part is getting pretty verbose that would be nice to clean up in the future.


justaugustus (approved):
Nice cleanup, @jeffmendoza!

PR #622, opened 2025-01-06
Update workflow policy to scan all branches for dangerous workflows #569

serb-google:
This PR addresses https://github.com/ossf/allstar/issues/569 and introduces 2 changes:
 - Adds support for local git repos in scorecard.go, with additional functionality to switch between all branches. I opted to not replace the existing Github client as policies may want to interact directly with Github.
 - Updates workflow.go to scan all remote branches in the local git repo.

Tested locally using https://github.com/serb-google/allstar, with a branch `insecure_test` which contains a fake unsafe workflow rule. Verified this triggered the alert which included output `Vulnerable Branch: refs/remotes/origin/insecure_test`.

 This is my first contribution to this repo, please review carefully :)

Issue #659, opened 2025-03-06
Release v4.3


coheigea:
Please consider releasing v4.3 so that features added (e.g. supporting globs - https://github.com/ossf/allstar/commit/4b1d42193ade50e7b21e56782530215bb9d86cdc) can be used in prod.

serb-google:
Tagging @jeffmendoza :)

jeffmendoza:
Yes, thanks for the nudge. Currently we have an issue
```
go: updates to go.mod needed; to update it:
	go mod tidy
```
Working through that with #688

justaugustus:
> Working through that with [#688](https://github.com/ossf/allstar/pull/688)

@jeffmendoza — I cleaned up and merged this. Also took care of most of the dependabot PRs.
Let's coordinate a time to do this soon. :)

PR #677, opened 2025-04-04
Add config option to specify branches for dangerous workflow

serb-google:
Add config option to specify branches for dangerous workflow, see https://github.com/ossf/allstar/pull/622#issuecomment-2740024164

Tested with and without option in `/.allstar/dangerous_workflow.yaml` file private branch to confirm this works as expected.

### DIFF
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -65,6 +65,11 @@ type OrgConfig struct {
 
 	// Schedule specifies whether to perform certain actions on specific days.
 	Schedule *ScheduleConfig `json:"schedule"`
+
+	// Comma-seperated branch list to scan for Dangerous Workflows.
+	// Blank/default to scan all branches.
+	// Must use format "refs/remotes/origin/branch_name".
+	DangerousWorkflowBranchList string `json:"dangerousWorkflowBranchList"`
 }
 
 // OrgOptConfig is used in Allstar and policy-specific org-level config to
@@ -108,6 +113,11 @@ type RepoConfig struct {
 
 	// Schedule specifies days during which to not send notifications,
 	Schedule *ScheduleConfig `json:"schedule"`
+
+	// Comma-seperated branch list to scan for Dangerous Workflows.
+	// Blank/default to scan all branches.
+	// Must use format "refs/remotes/origin/branch_name".
+	DangerousWorkflowBranchList string `json:"dangerousWorkflowBranchList"`
 }
 
 // RepoOptConfig is used in Allstar and policy-specific repo-level config to
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -66,7 +66,7 @@ type OrgConfig struct {
 	// Schedule specifies whether to perform certain actions on specific days.
 	Schedule *ScheduleConfig `json:"schedule"`
 
-	// Comma-seperated branch list to scan for Dangerous Workflows.
+	// Comma-separated branch list to scan for Dangerous Workflows.
 	// Blank/default to scan all branches.
 	// Must use format "refs/remotes/origin/branch_name".
 	DangerousWorkflowBranchList string `json:"dangerousWorkflowBranchList"`
@@ -114,7 +114,7 @@ type RepoConfig struct {
 	// Schedule specifies days during which to not send notifications,
 	Schedule *ScheduleConfig `json:"schedule"`
 
-	// Comma-seperated branch list to scan for Dangerous Workflows.
+	// Comma-separated branch list to scan for Dangerous Workflows.
 	// Blank/default to scan all branches.
 	// Must use format "refs/remotes/origin/branch_name".
 	DangerousWorkflowBranchList string `json:"dangerousWorkflowBranchList"`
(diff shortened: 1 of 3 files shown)


serb-google:
It's not really clear why the linter is failing, I can't replicate this locally with:
```
 allstar$ ../golangci-lint-1.64.7-linux-amd64/golangci-lint run --timeout 3m
 allstar$ ../golangci-lint-1.64.7-linux-amd64/golangci-lint cache clean
 ```

**Comment by serb-google** [comment]
Looks like https://github.com/ossf/allstar/pull/677/commits/24b42f006510ca3ab30c80c399685e8d8e0eb0cb resolved the lint error.

File: pkg/config/config.go [hunk 1]
```diff
@@ -65,6 +65,11 @@ type OrgConfig struct {
 
 	// Schedule specifies whether to perform certain actions on specific days.
 	Schedule *ScheduleConfig `json:"schedule"`
+
+	// Comma-seperated branch list to scan for Dangerous Workflows.
```
coheigea: Nit: separated. Same for below

coheigea (changes requested):

pkg/config/config.go
serb-google: Nice catch :)

coheigea:
What happens if a given branch isn't available in a repo? Will it just skip or throw an error? I'd want to specify "refs/remotes/origin/master", "refs/remotes/origin/main" etc. to catch different default branches in various repos. Or is there a way to always pick the default branch?

serb-google:
> What happens if a given branch isn't available in a repo? Will it just skip or throw an error? I'd want to specify "refs/remotes/origin/master", "refs/remotes/origin/main" etc. to catch different default branches in various repos. Or is there a way to always pick the default branch?

Based on https://github.com/serb-google/allstar/blob/config/pkg/policies/workflow/workflow.go#L150 this should throw an error. 

I think it should be possible to wire up `GetDefaultBranchName()` from the scorecard Git client for a magic string like `Default` ( https://github.com/ossf/scorecard/blob/main/clients/git/client.go#L316 ). I should be able to have that wired up in a few days.

pkg/config/config.go
```diff
@@ -65,6 +65,11 @@ type OrgConfig struct {
 
 	// Schedule specifies whether to perform certain actions on specific days.
 	Schedule *ScheduleConfig `json:"schedule"`
+
+	// Comma-separated branch list to scan for Dangerous Workflows.
+	// Blank/default to scan all branches.
+	// Must use format "refs/remotes/origin/branch_name".
+	DangerousWorkflowBranchList string `json:"dangerousWorkflowBranchList"`
```
jeffmendoza: I don't believe you need these config fields. Just the ones you added to the policy-specific config

pkg/policies/workflow/workflow.go
```diff
@@ -42,6 +43,11 @@ type OrgConfig struct {
 
 	// Action defines which action to take, default log, other: issue...
 	Action string `json:"action"`
+
+	// Comma-separated branch list to scan for Dangerous Workflows.
+	// Blank/default to scan all branches.
+	// Must use format "refs/remotes/origin/branch_name".
+	DangerousWorkflowBranchList string `json:"dangerousWorkflowBranchList"`
```
jeffmendoza: Because this is already a part of the Dangerous Workflow policy config, just name this "BranchList"

jeffmendoza (commented):
Looks good, some naming comments.
Lets preserve the ability to easily only check the default branch, as @coheigea mentioned.

pkg/policies/workflow/workflow.go
serb-google: Ack, updated!

pkg/config/config.go
serb-google: Ack, updated!

serb-google:
I modified the logic to add in a keyword "default" which will be replaced with the actual default git branch name.

I tested this using a private repo with configuration file `.allstar/dangerous_workflow.yaml` with the contents `branchList: default`. Log:
```
{"severity":"DEBUG","time":"2025-04-27T16:21:34Z","message":"local checkout -- installation access token retrieved"}
{"severity":"DEBUG","repo":"serb-google/test_private","time":"2025-04-27T16:21:34Z","message":"Repo checkout"}
{"severity":"DEBUG","time":"2025-04-27T16:21:34Z","message":"local checkout -- checkout success"}
{"severity":"DEBUG","branches":"refs/heads/main","time":"2025-04-27T16:21:34Z","message":"Running against branch list"}
{"severity":"DEBUG","branch":"refs/heads/main","time":"2025-04-27T16:21:34Z","message":"Branch checkout"}
{"severity":"INFO","org":"serb-google","repo":"test_private","area":"Dangerous Workflow","result":true,"enabled":false,"notify":"","details":{"Findings":null},"time":"2025-04-27T16:21:34Z","message":"Policy run result."}
```

jeffmendoza (approved):
Added a quick note on how the lists are merged. Thanks for the updates!!

jeffmendoza:
Some merge conflicts for some reason, a rebase was needed, will merge now

coheigea:
@jeffmendoza Can we have a new release with this feature please? https://github.com/ossf/allstar/issues/659

PR #697, opened 2025-06-06
Don't error out when unable to comment.

jeffmendoza:
Sometimes the issue is found, but Allstar is unable to comment. Don't exit out of the enforce loop on this, just log a Warning and continue.

The error I am currently seeing is "403 Commenting is disabled on issues with more than 2500 comments"

### DIFF
--- a/pkg/issue/issue.go
+++ b/pkg/issue/issue.go
@@ -139,9 +139,17 @@ func ensure(ctx context.Context, c *github.Client, issues issues, owner, repo, p
 	if !strings.Contains(issue.GetBody(), hash) && hasIssueSection(issue.GetBody(), updateSectionName) {
 		// Comment update and update issue body
 		commentBody := fmt.Sprintf("The policy result has been updated.\n\n---\n\n%s", text)
-		comment, _, err := issues.CreateComment(ctx, owner, issueRepo, issue.GetNumber(), &github.IssueComment{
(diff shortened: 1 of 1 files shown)


pkg/issue/issue.go
```diff
@@ -139,9 +139,17 @@ func ensure(ctx context.Context, c *github.Client, issues issues, owner, repo, p
 	if !strings.Contains(issue.GetBody(), hash) && hasIssueSection(issue.GetBody(), updateSectionName) {
 		// Comment update and update issue body
 		commentBody := fmt.Sprintf("The policy result has been updated.\n\n---\n\n%s", text)
-		comment, _, err := issues.CreateComment(ctx, owner, issueRepo, issue.GetNumber(), &github.IssueComment{
+		comment, rsp, err := issues.CreateComment(ctx, owner, issueRepo, issue.GetNumber(), &github.IssueComment{
 			Body: &commentBody,
 		})
+		if err != nil && rsp != nil && (rsp.StatusCode == http.StatusGone || rsp.StatusCode == http.StatusForbidden) {
+			log.Warn().
```
spencerschrock, line 146: not familiar with what you're seeing in the logs, but will err always be non-nil for these status codes?

spencerschrock (approved):

pkg/issue/issue.go
jeffmendoza: Yeah, the GitHub client is unlike the stdlib http client, it sets err when any response is not 200.

PR #728, opened 2025-08-22
general housekeeping

cpanato:
- format .md files and clean up 
- fix shell issue and update shebang 

cc @justaugustus

### DIFF
--- a/MAINTAINERS.md
+++ b/MAINTAINERS.md
@@ -20,4 +20,4 @@
 Former maintainers are listed here.
 Thanks for your contributions to Scorecard!
 
-- 
+-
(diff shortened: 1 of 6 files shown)


cloudbuild.yaml
```diff
@@ -1,7 +1,7 @@
 steps:
 - name: golang:1.24
   entrypoint: go
-  args: ['install', 'github.com/google/ko@v0.18.0']
+  args: ['install', 'github.com/ko-build/ko@v0.18.0']
```
jeffmendoza: This doesn't work, still declared as "google": https://github.com/ko-build/ko/blob/main/go.mod#L1

Not sure on the latest: https://github.com/ko-build/ko/issues/791

cpanato: you right i thought i already fixed that, will drop this for now

jeffmendoza (approved):

PR #841, opened 2026-05-28
feat: add Secret Scanning policy

bmendonca3:
## Summary

Adds a Secret Scanning policy for Allstar.

This policy:
- uses `secret_scanning.yaml` for org/repo configuration
- checks whether GitHub reports repository secret scanning as enabled
- reports a violation only when the setting is explicitly `disabled`
- treats missing or unavailable `security_and_analysis` data as unavailable, not as disabled
- supports the `fix` action by enabling secret scanning when GitHub reports it as disabled
- documents the policy in the README

This intentionally leaves push protection out of scope for the first pass.

Fixes #363

## Testing

- `PATH=/tmp/go/bin:$PATH go test ./pkg/policies/secretscanning`
- `PATH=/tmp/go/bin:$PATH go test ./pkg/policies/...`
- `PATH=/tmp/go/bin:$PATH go test ./...`
- `PATH=/tmp/go/bin:$PATH go build ./...`
- `PATH=/tmp/go/bin:$PATH go vet ./...`
- `PATH=/tmp/go/bin:$PATH /tmp/go-tools/golangci-lint run --timeout 5m --verbose`

## Live validation

Validated the policy-specific Check/Fix/Check path against a disposable repository:
https://github.com/bmendonca3/allstar-secret-scanning-validation-20260528152234

Using an authenticated GitHub client against this branch, the policy:
- observed GitHub API `secret_scanning.status=disabled`
- returned `Check` result `enabled=true pass=false`
- applied `Fix`
- observed GitHub API `secret_scanning.status=enabled`
- returned follow-up `Check` result `enabled=true pass=true`

This validates the live GitHub API Check/Fix/Check path for this policy. It does not claim full end-to-end GitHub App installation or issue-creation coverage, which remains outside this validation pass.

PR #842, opened 2026-05-28
feat: recognize rulesets in branch protection

### LINKED ISSUE #475: [Feature Request]: Check rulesets for branch protection rules
Currently it appears that allstar is not checking the GitHub rulesets active on a repository. Rulesets can be used to set branch protection rules on a repository and the details of it are obtained via a different API endpoint to the standard branch protection rules. This means that currently allstar can fail on branch protection rules, even when they've been implemented via rulesets.

As a user 
I would like allstar to check for branch protection rules from both sets of data
So that I am not alerted for failing to implement branch protection when it's been implemented by a ruleset

As mentioned by @jeffmendoza I think combining the rulesets with branch protections policy makes sense as they do the same thing just in different ways.

Thanks :)

Initial question to investigate lack of checking regarding rulesets:
> Hey 👋🏻 

> I've started to implement branch protection rules on repos via [GitHub rulesets](https://docs.github.com/en/repositories/configuring-branches-and-merges-in-your-repository/managing-rulesets/about-rulesets#about-rulesets-protected-branches-and-protected-tags), rather than the original way of setting the protection and allstar has started raising issues saying they don't exist. 

> Is that because allstar isn't checking for rulesets or because I'm still missing something? I did have a little read of the repo and couldn't find specific reference. If it's not checking rulesets, will this be implemented in future?

bmendonca3:
## Summary

Adds Check-only Repository Rulesets support to the existing Branch Protection policy.

This keeps the existing classic branch protection behavior, but when a branch is missing or incomplete under classic branch protection, Allstar also checks the active rules returned by `Repositories.GetRulesForBranch`.

The policy now recognizes direct ruleset equivalents for:
- pull request reviews
- required approving review count
- stale review dismissal
- code owner reviews
- required status checks
- strict/up-to-date status checks
- signed commits
- non-fast-forward protection

This intentionally does not add Fix/CRUD support for rulesets. The existing `fix` action still updates classic branch protection only.

Fixes #475

## Testing

- `PATH=/tmp/go/bin:$PATH go test ./pkg/policies/branch`
- `PATH=/tmp/go/bin:$PATH go test ./pkg/policies/...`
- `PATH=/tmp/go/bin:$PATH go test ./...`
- `PATH=/tmp/go/bin:$PATH go vet ./...`
- `PATH=/tmp/go/bin:$PATH go build ./...`
- `PATH=/tmp/go/bin:$PATH /tmp/go-tools/golangci-lint run --timeout 5m --verbose`
- `git diff --check`

## Live validation

Validated against a disposable public repository:
https://github.com/bmendonca3/allstar-rulesets-validation-20260528154925

Before creating a ruleset:
- active branch rules: `pull_request=0 status_checks=0 non_fast_forward=0 required_signatures=0`
- Allstar Check result: `enabled=true pass=false`
- notify text: `No protection found for branch main`

After creating an active ruleset for `main` with pull request review and non-fast-forward rules:
- active branch rules: `pull_request=1 status_checks=0 non_fast_forward=1 required_signatures=0`
- Allstar Check result: `enabled=true pass=true`
- notify text: empty

## Scope note

Rulesets are not used to satisfy `enforceOnAdmins`, because rulesets model bypass behavior differently than classic branch protection.

### DIFF
--- a/README.md
+++ b/README.md
@@ -321,12 +321,19 @@ here](https://pkg.go.dev/github.com/ossf/allstar/pkg/policies/branch#OrgConfig).
 
 The branch protection policy checks that GitHub's [branch protection
 settings](https://docs.github.com/en/github/administering-a-repository/defining-the-mergeability-of-pull-requests/about-protected-branches)
-are setup correctly according to the specified configuration. The issue text
-will describe which setting is incorrect. See [GitHub's
+or active repository rulesets are setup correctly according to the specified
+configuration. The issue text will describe which setting is incorrect. See
+[GitHub's
 documentation](https://docs.github.com/en/github/administering-a-repository/defining-the-mergeability-of-pull-requests/about-protected-branches)
 for correcting settings.
 
-The `fix` action will change the branch protection settings to be in compliance with the specified policy configuration.
+Rulesets are checked for requirements that directly map to branch protection
+configuration, such as pull request reviews, required status checks, signed
+commits, and non-fast-forward protection. Rulesets are not used to satisfy
+`enforceOnAdmins`, because rulesets model bypass differently than classic
+branch protection. The `fix` action changes classic branch protection settings
+to be in compliance with the specified policy configuration; it does not create
+or update rulesets.
 
 ### Binary Artifacts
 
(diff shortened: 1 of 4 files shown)


justaugustus (commented):
See note about AI contributions —> https://github.com/ossf/allstar/pull/841#issuecomment-4569938183

PR #845, opened 2026-06-02
enforce: warn when app has zero installations

### LINKED ISSUE #294: App doesn't report an error when it exists, but is not installed in any organisations
I had created a GitHub Application in our private GitHub enterprise organisation, using the steps described at the bottom of this issue, and [here](https://github.com/ossf/allstar/blob/main/operator.md#create-a-github-app).

I was quite perplexed as to why it wasn't applying any configurations. We were seeing repeated messages like the below on each configured loop interval, so it looked to us that it must have authenticated OK.

```
{"severity":"INFO","area":"bot","count":0,"time":"2022-09-21T13:56:53+10:00","message":"Enforcing policies on installations."}
{"severity":"INFO","area":"bot","size":0,"items":0,"time":"2022-09-21T13:56:53+10:00","message":"Total cache size."}
{"severity":"INFO","area":"bot","count":0,"results":{},"time":"2022-09-21T13:56:53+10:00","message":"EnforceAll complete."}
```

As a sanity test, we verified that we received an error when using an invalid `PRIVATE_KEY`:
```
{"severity":"ERROR","error":"could not parse private key: invalid key: Key must be a PEM encoded PKCS1 or PKCS8 key","time":"2022-09-21T14:04:42+10:00","message":"Unexpected error enforcing policies."}
```

and we also received an error when running with an invalid `APP_ID`:

```
{"severity":"ERROR","error":"GET https://api.github.com/app/installations?per_page=100: 401 Integration must generate a public key []","time":"2022-09-21T14:00:36+10:00","message":"Unexpected error enforcing policies."}
```

I realised that I needed to install the app into an organisation, in order for it to work. This step is missing in the docs linked at the top of this issue.

It would be useful to update the docs to add a step to install the app into an org (via the applications Settings page -> Install App tab), and also to log a message, e.g. "App authentication succeeded but no Installations were found", to direct people like me towards installing the app into the relevant orgs.

For completeness / because I was writing this up anyway, I thought I'd share the steps used to build and run this ourselves:

### Installation steps
1. Create and configure a new GitHub app in the org, collecting the AppID and Private Key.
2. Install the app into the required organisation(s)
3. Add the AppID to an `APP_ID` environment variable on your system
4. Add the Private Key to a `PRIVATE_KEY` environment variable on your system (something like the below will work):

```
export PRIVATE_KEY=`cat /tmp/allstar-private-key.pem`
```

5. Set a `KEY_SECRET` environment variable to `direct`
6. Clone the Allstar repository `git clone git@github.com:ossf/allstar.git`
7. Run `go build cmd/allstar/main.go`
8. Run `go run cmd/allstar/main.go`

arpitjain099:
Fixes #294

### Problem

When the Allstar GitHub App is authenticated but installed in zero organizations, `EnforceAll` (in `pkg/enforce/enforce.go`) lists installations, logs `count:0` and `EnforceAll complete` at INFO level, and then iterates an empty slice without enforcing anything. There is no signal that the app simply was not installed anywhere, which is the most common cause. The issue reporter spent time debugging this because invalid `PRIVATE_KEY` / `APP_ID` both produce clear errors, but the much more likely "authenticated but not installed" case is silent.

### Change

After the existing `Enforcing policies on installations.` log, when `len(insts) == 0`, emit a `log.Warn` stating that authentication succeeded but no installations were found, that no policies will be enforced, and pointing the operator at the install step (app Settings page -> Install App tab). The logging matches the surrounding zerolog style (`Str("area", "bot")`).

### Test

Added `TestEnforceAllNoInstallations`, which mocks `getAppInstallations` to return an empty slice, captures the zerolog global logger into a buffer, and asserts a warn-level entry mentioning "no installations were found" is produced. It also confirms `getAppInstallationRepos` is not called and the results map is empty. I verified the test is a real regression test: it fails against the unpatched `enforce.go` and passes with the fix.

```
$ go test ./pkg/enforce/...
ok  	github.com/ossf/allstar/pkg/enforce	1.296s
```

`gofmt -l` and `go vet ./pkg/enforce/...` are both clean on the changed files.

### Note on docs

The original issue also suggested adding a step to `operator.md` documenting that the app must be installed into an org. I kept this PR scoped to the code fix and its test so the behavior change is easy to review on its own. Happy to follow up with the docs update (or fold it in here) if maintainers prefer.

This commit is DCO signed off (`Signed-off-by`).

### DIFF
--- a/pkg/enforce/enforce.go
+++ b/pkg/enforce/enforce.go
@@ -89,6 +89,15 @@ func EnforceAll(ctx context.Context, ghc ghclients.GhClientsInterface, specificP
 		Int("count", len(insts)).
 		Msg("Enforcing policies on installations.")
 
+	if len(insts) == 0 {
+		log.Warn().
+			Str("area", "bot").
+			Msg("App authentication succeeded but no installations were found. " +
+				"Allstar will not enforce any policies. The app is likely not " +
+				"installed on any organization. Install it on the relevant org(s) " +
+				"via the app's Settings page -> Install App tab.")
+	}
+
 	g, ctx := errgroup.WithContext(ctx)
 	g.SetLimit(operator.NumWorkers)
 	var mu sync.Mutex
--- a/pkg/enforce/enforce_test.go
+++ b/pkg/enforce/enforce_test.go
@@ -15,6 +15,7 @@
 package enforce
 
 import (
+	"bytes"
 	"context"
 	"encoding/binary"
 	"errors"
@@ -25,6 +26,8 @@ import (
 
 	"github.com/google/go-cmp/cmp"
 	"github.com/google/go-github/v84/github"
+	"github.com/rs/zerolog"
+	"github.com/rs/zerolog/log"
 
 	"github.com/ossf/allstar/pkg/config/operator"
 	"github.com/ossf/allstar/pkg/policydef"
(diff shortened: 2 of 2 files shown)


arpitjain099:
Still active on this. Let me know if there are any changes you'd like or if a maintainer can take a look when they get a chance.

PR #864, opened 2026-07-26
Handle 404 from UpdateBranchProtection in branch policy Fix

### LINKED ISSUE #562: Issue with Allstar Branch Protection Enforcement (404 errors upon action: fix)
Hello,

Over the past couple of months, I've noticed a recurring issue where Allstar returns 404 error messages when attempting to enforce branch protection, as shown in the logs below, causing the application to crash. The issue is still there with Allstar latest version (4.2)

```bash
allstar {"severity":"INFO","org":"Org","repo":"test0","area":"bot","enabled":true,"time":"2024-08-20T08:29:17Z","message":"Check repo enabled"}                                                                                                  
allstar {"severity":"INFO","org":"Org","repo":"test0","area":"Binary Artifacts","enabled":false,"time":"2024-08-20T08:29:17Z","message":"Check repo enabled"}                                                                                    
allstar {"severity":"INFO","org":"Org","repo":"test0","area":"Binary Artifacts","result":true,"enabled":false,"notify":"","details":{"Artifacts":null},"time":"2024-08-20T08:29:17Z","message":"Policy run result."}                             
allstar {"severity":"INFO","org":"Org","repo":"test0","area":"Branch Protection","enabled":true,"time":"2024-08-20T08:29:17Z","message":"Check repo enabled"}                                                                                    
allstar {"severity":"INFO","org":"Org","repo":"test0","area":"Branch Protection","result":false,"enabled":true,"notify":"No protection found for branch master\n","details":{"master":{"PRReviews":false,"NumReviews":0,"DismissStale":false,"Bl 
allstar {"severity":"ERROR","error":"PUT [https://api.github.com/repos/Org/test0/branches/master/protection](https://api.github.com/repos/Org/test0/branches/master/protection): 404 Branch protection has been disabled on this repository. []","time":"2024-08-20T08:29:18Z","message":"Unexpected error running policies."}
allstar {"severity":"INFO","area":"bot","count":7,"results":{"CODEOWNERS":{"totalFailed":1},"OpenSSF Scorecard":{"totalFailed":1},"Repository Administrators":{"totalFailed":1}},"time":"2024-08-20T08:29:18Z","message":"EnforceAll complete."}           
```

After investigating, I found that this issue occurs when branch protection rules are disabled on a repository. I also noticed in the GitHub API documentation that while there is an [API to update branch protection](https://docs.github.com/en/rest/branches/branch-protection?apiVersion=2022-11-28#get-branch-protection), there doesn’t seem to be one to set it initially. Could this be a new limitation from GitHub?

Screenshot 2024-08-20 at 10 34 55

Currently, I’ve implemented a custom alerting system to manually re-enable branch protection rules when Allstar fails, but this is not an ideal solution. Is there a better workaround that you would recommend?

For reference, here are the current permissions for my Allstar GitHub App.

Screenshot 2024-08-20 at 10 33 39

pujitha24:
Motivation:
Allstar's Branch Protection policy fails to enforce branch protection
on repos where branch protection has been disabled (e.g. via GitHub's
rulesets migration). GitHub's PUT
/repos/{owner}/{repo}/branches/{branch}/protection endpoint returns a
404 ("Branch protection has been disabled on this repository") in
this case, which fix() previously propagated as an error. That error
bubbles up through runPolicies() to runPoliciesOnInstRepos(), which
breaks out of its loop over repos on the first error. Net effect: once
one repo in a GitHub App installation hits this 404, every other repo
in that installation is skipped for the rest of that enforcement
cycle, every cycle, until the affected repo is fixed or removed. This
does not crash the process (EnforceAll logs the error and returns nil
from the errgroup), so no outage occurs, but enforcement silently
stops for unrelated repos in the same installation.

Approach:
Mirror the existing http.StatusForbidden handling already present at
both call sites of rep.UpdateBranchProtection() in fix(): when the
response status is 404, log a Warning (matching the existing log
style/fields) and return nil instead of propagating the error, so the
enforcement loop for other repos/installations is unaffected. This
matches the short-term fix suggested by maintainer jeffmendoza on the
issue: "update the code to expect the 404 on some repos, and just log
a Warning and continue without exiting the enforcement loop."

Validation:
Added two new TestFix subtests in pkg/policies/branch/branch_test.go
that simulate a 404 from UpdateBranchProtection on both the
create-from-scratch path and the update-existing-protection path, and
assert fix() returns nil without issuing any protection request.
Verified both new subtests fail with the pre-fix code (via `git stash`
on branch.go) and pass with the fix applied.

  go build ./...
  go test ./pkg/policies/branch/...
  ok  	github.com/ossf/allstar/pkg/policies/branch	0.251s

Fixes #562

Signed-off-by: Pujitha Paladugu <10557236+pujitha24@users.noreply.github.com>

### DIFF
--- a/pkg/policies/branch/branch.go
+++ b/pkg/policies/branch/branch.go
@@ -494,6 +494,15 @@ func fix(ctx context.Context, rep repositories, c *github.Client,
 						// no sense to continue, just return
 						return nil
 					}
+					if rsp != nil && rsp.StatusCode == http.StatusNotFound {
+						log.Warn().
+							Str("org", owner).
+							Str("repo", repo).
+							Str("area", polName).
+							Msg("Fix action selected, but branch protection has been disabled on this repository.")
+						// no sense to continue, just return
+						return nil
+					}
 					return err
 				}
 				continue
@@ -634,6 +643,14 @@ func fix(ctx context.Context, rep repositories, c *github.Client,
 						Str("repo", repo).
 						Str("area", polName).
 						Msg("Action set to fix, but did not accept admin:write permissions update.")
+					return nil
+				}
+				if rsp != nil && rsp.StatusCode == http.StatusNotFound {
+					log.Warn().
+						Str("org", owner).
+						Str("repo", repo).
+						Str("area", polName).
+						Msg("Fix action selected, but branch protection has been disabled on this repository.")
 					return nil
 				}
 				return err
--- a/pkg/policies/branch/branch_test.go
+++ b/pkg/policies/branch/branch_test.go
@@ -1371,6 +1371,7 @@ func TestFix(t *testing.T) {
 		Prot                 map[string]github.Protection
 		SignatureProt        map[string]github.SignaturesProtectedBranch
 		cofigEnabled         bool
+		FailUpdateWith       int
 		Exp                  map[string]github.ProtectionRequest
 		ExpSignatureRequests map[string]bool
 	}{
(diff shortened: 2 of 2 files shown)


pujitha24:
This is rebased and green whenever you get a chance to take a look — happy to make any changes that would help review go smoother.

PR #872, opened 2026-08-13
Retry transient GitHub gateway errors

### LINKED ISSUE #57: Add retry to transport
We get 502 and 504 occasionally from GitHub. These should be retried.

CAOShurong:
## Summary

- retry GitHub API responses with status 502 or 504 using exponential backoff
- limit automatic retries to idempotent HTTP methods so issue/comment creation and other non-idempotent mutations are not duplicated
- preserve the final gateway response when the retry budget is exhausted

The implementation reuses `go-retryablehttp`, which was already present in the module graph, and applies the retry transport to both app-level and installation-level GitHub clients.

Fixes #57.

## Validation

- `go test ./pkg/ghclients -count=1`
- real `httptest` + `go-github` paths: 502 -> 504 -> 200, non-gateway 500, non-idempotent POST, and persistent 504
- `go test` for every package except `pkg/scorecard`
- `go build ./...`
- `go vet ./...`
- `go mod verify`
- `golangci-lint run --timeout 5m --new-from-rev=HEAD`

On Windows, the unmodified `pkg/scorecard` `TestLocal` fails after its temporary `go-git/go-git` clone reports unstaged changes while switching branches. All other packages pass; this change does not touch that package.

## AI assistance

OpenAI Codex assisted with issue screening, implementation, and tests. I reviewed the patch and ran the validation listed above.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -11,6 +11,7 @@ require (
 	github.com/google/go-cmp v0.7.0
 	github.com/google/go-github/v84 v84.0.0
 	github.com/gregjones/httpcache v0.0.0-20190611155906-901d90724c79
+	github.com/hashicorp/go-retryablehttp v0.7.8
 	github.com/ossf/scorecard/v5 v5.5.0
 	github.com/rhysd/actionlint v1.7.12
 	github.com/rs/zerolog v1.35.1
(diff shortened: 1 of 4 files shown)


CAOShurong:
Not stale — still ready for review. The branch is up to date and mergeable; no rebase needed. Happy to address any feedback on the retry implementation (pkg/ghclients/retry.go) if anything needs adjusting.

PR #880, opened 2026-08-31
Remove OpenSSF-hosted app installation path

justaugustus:
## Why

OpenSSF Scorecard's hosted infrastructure is moving off GCP to AWS operated by
the AboutCode Foundation, because foundation GCP funding has ended
(ossf/scorecard#5208). The Allstar hosted GitHub App ran on that same
infrastructure and is retired with it.

That leaves this repository's documentation wrong in a way that matters.
Quickstart and Manual Installation both open by telling the reader to install
`github.com/apps/allstar-app`, and the self-hosting section is written to talk
them out of it ("Only self-host if you must!"). Someone following the README
today ends up with an app that will never act on their organization. The
compiled-in operator defaults have the same problem: they point at the hosted
instance's App ID and at its Secret Manager path in the `allstar-ossf` GCP
project.

ref: https://github.com/ossf/allstar/issues/881, ossf/scorecard#5208

## Outcome

- Installation is restructured around what is now true. Every deployment needs
  a GitHub App and an `.allstar` control repository, and then picks the GitHub
  Action or the service daemon. The two are presented as peers rather than
  ranked by effort against an option that no longer exists.
- A migration section answers the question an affected organization will
  actually ask — whether they lose their configuration. They do not.
- Operator defaults no longer reference the OpenSSF instance, and Allstar exits
  at startup with an actionable error when `APP_ID` or the private key is
  missing, rather than attempting to authenticate as an app the operator does
  not own and surfacing GitHub API errors that point nowhere near the mistake.
- The GCP App Engine deployment files are removed. The sizing and the two
  non-default settings the hosted instance ran with are recorded in
  `operator.md` first, since that is the part worth keeping.

## Scope

- **GCP Secret Manager remains supported.** The `gocloud.dev` runtimevar driver
  stays registered; only the hardcoded path to OpenSSF's own secret is removed.
  Self-hosters using GCP are unaffected.
- **The Cloud Logging severity mapping in `setupLog()` is deliberately kept.**
  Changing those field names would break log parsing for existing deployments.
  The comment is reworded; the behavior is not.
- Policy behavior, the config schema, and the `.allstar` format are untouched.
  This changes how Allstar is obtained and configured, not what it does.
- `install-decision.jpg` is replaced by an equivalent Mermaid flowchart. The
  decision tree itself is unchanged, but no source file for the image is
  checked in, so it could not be edited.

## Verification

Beyond what CI covers: built the binary and confirmed both new startup failures
fire with the intended message — unset `APP_ID`, and `KEY_SECRET=direct` with no
`PRIVATE_KEY`.

Also checked that every intra-document anchor in the Markdown resolves. That
turned up three links already broken on `main` — `README#operator-instructions`,
`github-action-installation#private-key`, and an `opt-out.md` link missing its
`#` — each in a file this PR was already editing, so they are fixed here.

## Review focus

- **The retirement notice says "has been retired" with no date.** If the hosted
  app is still serving when this merges, that wording needs to change.
- **The claim that migration is lossless.** It rests on `getPolicyIssue`
  matching on label and title with `State: "all"` and no author filter, so a
  newly created app adopts and closes the hosted app's existing issues instead
  of filing duplicates. Worth a second reader.
- **`KEY_SECRET` now defaults to `direct`,** which reads the key from the
  environment. It is the only backend that works without the operator first
  provisioning external infrastructure, and it is what the Action deployment
  already used — but it is a weaker default than a secret store, so it is a
  deliberate trade worth confirming.


README.md
```diff
@@ -4,6 +4,19 @@
...
 
+> [!IMPORTANT]
+> **The OpenSSF-hosted Allstar app has been retired.** It ran on Google Cloud
+> Platform infrastructure whose funding has ended, and installing
+> `github.com/apps/allstar-app` no longer does anything. Allstar itself is
+> unaffected and continues to be maintained — you now run it yourself, either
+> [as a GitHub Action](#running-allstar-as-a-github-action) or
+> [as a service daemon](#running-allstar-as-a-service-daemon). See
+> [ossf/scorecard#5208](https://github.com/ossf/scorecard/issues/5208) for
+> background on the infrastructure migration.
```
justaugustus:
```suggestion
> [!IMPORTANT]
> **The OpenSSF-hosted Allstar GitHub App has been retired.** 
> Allstar, the OpenSSF Scorecard subproject, itself continues to be
> maintained — you must now run it yourself, either
> [as a GitHub Action](#running-allstar-as-a-github-action) or
> [as a service daemon](#running-allstar-as-a-service-daemon).
>
> See [ossf/allstar#881](https://github.com/ossf/allstar/issues/881) for
> more details.
```

PR #893, opened 2026-09-18
fix(branch): keep enforcement going when branch protection is unavailable

### LINKED ISSUE #562: Issue with Allstar Branch Protection Enforcement (404 errors upon action: fix)

Eljees:
Fixes #562.

### The failure

`PUT /repos/{owner}/{repo}/branches/{branch}/protection` answers `404` when the setting is not there to be written. Two ordinary cases produce it: a branch listed in `enforceBranches` that does not exist, and a repository whose protection has been migrated to rulesets, where the classic endpoint is gone. `GetBranchProtection` also returns `404` for such a branch, so `fix()` takes the "no existing protection, create from config" path and then hits the `404` on the write.

Only `403` was handled there, so the `404` was returned as an error. That error leaves `fix()`, and with it the enforcement run: every repository queued behind this one in the same App installation is skipped, which is what #562 reports. A single repository that moved to rulesets stops enforcement for the whole installation.

### The change

`404` is now treated the way the same status is already treated for commit signing (`getSignatureProtectionEnabled`, which has returned "not enabled" on `404` since it was written): log a warning naming the branch and carry on. `403` handling is untouched, and any other status is still returned as an error.

Three call sites reach the classic protection endpoints per branch, and all three had the same gap:

* `UpdateBranchProtection` on the create-from-config path,
* `UpdateBranchProtection` on the update-existing path,
* `RequireSignaturesOnProtectedBranch`, which is called per branch right after, on protection that is not there either.

The warning says the protection is not available for that branch rather than naming rulesets as the cause, since `404` does not tell us which of the two cases it was.

`continue` rather than `return nil`: a branch that cannot take protection says nothing about the next branch in `enforceBranches`. A repository on rulesets will simply skip all of them in turn.

### Tests

`pkg/policies/branch/branch_rulesets_test.go`:

* `TestFixContinuesWhenBranchProtectionIsNotAvailable` — the first branch answers `404` on the write, and the test asserts the run reached the *following* branch. Asserting only that `Fix` returned no error would also pass if the loop had quietly stopped after the first branch, so the assertion is on the call that has to happen, not on the absence of an error.
* `TestFixContinuesWhenSignedCommitsCannotBeRequired` — the same for the signed-commits call.
* `TestFixStillFailsOnOtherErrors` — a `500` is still reported, so a broken installation does not start looking healthy.

I checked the failure direction by reverting only `branch.go` and keeping the tests: both continuation tests fail with `404 Not Found`, the `500` test still passes, and the rest of the package is unaffected. `gofmt`, `go vet` and `go test ./pkg/policies/branch/` are clean on Go 1.26.7.

### Relation to #864

@pujitha24 opened #864 for the same issue in July; it was closed by the stale bot last week without a human review. This change differs in three ways, two of them from the automated review left on that PR: it continues to the next branch instead of returning, it does not state rulesets as the cause in the log message, and it also covers `RequireSignaturesOnProtectedBranch`, which is the third place the same `404` arrives.


Eljees:
The red `deadcode` check here is not about this change: the job fails while installing the tool, before it ever looks at the repository.

```
go install golang.org/x/tools/cmd/deadcode@latest
go: golang.org/x/tools/cmd/deadcode@latest: golang.org/x/tools@v0.50.0
    requires go >= 1.26.0 (running go 1.25.9; GOTOOLCHAIN=local)
```

`golang.org/x/tools` v0.50.0 was published on 8 September and raised its own Go requirement to 1.26, while the step resolves `@latest` with `GOTOOLCHAIN=local` on the toolchain that matches `go.mod` (1.25.9), so the install cannot pick a newer toolchain. The same failure is on other open pull requests that touch nothing related - #892 and #889, for instance.

Pinning the tool to `golang.org/x/tools/cmd/deadcode@v0.49.0`, or letting that step use `GOTOOLCHAIN=auto`, would make it run again. Happy to send that as a separate pull request if it would be useful - it is unrelated to this fix, so I have left it out here.

Everything else is green, and `gofmt`, `go vet` and `go test ./pkg/policies/branch/` pass locally on Go 1.26.7.

Eljees:
Sent as offered above: #894 puts `GOTOOLCHAIN=auto` on that one install step. Its own `deadcode` check came back green, the first one since 8 September, so the red here belongs to the job rather than to this branch. All eleven checks pass there, `zizmor` included.

